PR #7, opened 2020-05-29
Monaco editor

StefanRetief:
Added @bentley/monaco-editor to the project. The package allows for live code viewing via a button on the left side of the screen. As the user switches through the different samples, the code viewer will also update with the code related to that sample. The code editor is size adjustable, with the minimum width being 500px.

Note the changes to the SampleSpec interface: Each sample spec should also include the files to load in the code viewer, which will be loaded via a raw-loader. Files that are provided to the code viewer should include a name, without any slashes or indication of directory and an import statement that is **relative to the sample spec's location.**

Note that when each file is loaded into the code viewer, each of the relative imports in the code provided will be unnested.

**Known Issues:**
1. If the user switches between multiple samples quickly, the application crashes. This issue is related to validation occurring within the editor package.
2. On occasion, when switching the samples, an error message appears in the console related to not finding a file related to the sample. This issue is also related to code model handling within the editor package.

**Related changes:** This Pull request has several changes related to the sample spec for the existing samples. All sample specs have been moved from the index.tsx file in each sample, into their own sampleSpec.tsx file. **50 out of 59 file changes are related to moving the sample spec**


StefanRetief:
@BaoTon @Josh-Schifter @Ellord207 Could you sign the CLA for this branch? Hoping to get this PR in before it gets too stale

PR #11, opened 2020-06-04
Use dummy sample showcase user for no-signin access.

BaoTon:
The PR uses our sample showcase user provided by our imodel.js developer services. This provides 3rd party developers a way to run the sample without requiring a login or access to the projects we use in the showcase.

### DIFF
--- a/src/NoSignInIAuthClient.ts
+++ b/src/NoSignInIAuthClient.ts
@@ -1,0 +1,57 @@
+import { AuthStatus, BeEvent, BentleyError, ClientRequestContext } from "@bentley/bentleyjs-core";
+import { AccessToken } from "@bentley/itwin-client";
+import { FrontendAuthorizationClient } from "@bentley/frontend-authorization-client";
+
+export class NoSignInIAuthClient implements FrontendAuthorizationClient {
+  public readonly onUserStateChanged: BeEvent<(token: AccessToken | undefined) => void>;
+  protected _accessToken?: AccessToken;
+
+  constructor() {
+    this.onUserStateChanged = new BeEvent();
+  }
+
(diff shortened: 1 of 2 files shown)


src/NoSignInIAuthClient.ts
```diff
@@ -0,0 +1,57 @@
+import { AuthStatus, BeEvent, BentleyError, ClientRequestContext } from "@bentley/bentleyjs-core";
```
calebmshafer, line 1: copyright header

src/NoSignInIAuthClient.ts
```diff
@@ -0,0 +1,57 @@
...
+  public get hasSignedIn(): boolean {
+    return !!this._accessToken;
+  }
+
+  public async generateTokenString(requestContext?: ClientRequestContext) {
+    if (requestContext) {
+      requestContext.enter();
+    }
+
+    const response = await fetch(`https://prod-imodeldeveloperservices-eus.azurewebsites.net/api/v0/sampleShowcaseUser`);
```
calebmshafer, line 43: Can we make this an environment variable?  It will get bundled at build time then...

BaoTon: Yeah I think the URL makes sense to be set as an environment variable - let me make the necessary modifications.

BaoTon:
Do we not have an .env/config file to easily opt-in/opt-out of options? Would this be a good idea instead of editing code?

Josh-Schifter:
> Do we not have an .env/config file to easily opt-in/opt-out of options? Would this be a good idea instead of editing code?

I don't expect anyone to build this themselves so we don't have to make it easy to switch back and forth at build time like that.  Eventually, I'm hoping we can switch _at runtime_ when the user chooses 'browse to my own iModel'.

PR #25, opened 2020-06-18
Enable skybox for 3d  views.

pmconne:


pmconne:
@Josh-Schifter because it's 2020 and we are *still* dealing with different line endings on unix and windows. You can change your settings to tell it to ignore whitespace changes.
image

pmconne:
@Josh-Schifter Ugly brown is because you're in a top view in a non-geolocated file. Rotate the view to see why. If you prefer, I can override the default skybox settings.

Josh-Schifter:
> @Josh-Schifter Ugly brown is because you're in a top view in a non-geolocated file. Rotate the view to see why. If you prefer, I can override the default skybox settings.

I don't think the ugly brown affected Retail Building (maybe I missed it).  I think we should do something because it looks pretty bad.  Either like you say, override the default skybox, or only to skybox for geolocated, or only do skybox for certain samples.

pmconne:
Ugly brown is the default color you see when looking straight down in a top view of any non-geolocated model, including the retail building.
I changed that color. Is it less ugly now?

Josh-Schifter:
> I changed that color. Is it less ugly now?

Seems like a good compromise.

PR #28, opened 2020-06-23
More tree samples

tomdicarlo:
Added the remaining tree samples. Added a generic reloadable component.


src/Components/ReloadableComponent/ReloadableConnection.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export interface ReloadableConnectionProps {
+  iModelName: string;
+  onIModelReady?: (imodel: IModelConnection) => void;
+}
+
+export interface ReloadableConnectionState {
+  /** iModel whose contents should be displayed in the viewport */
+  imodel?: IModelConnection;
+  /** View state to use when the viewport is first loaded */
```
Josh-Schifter, line 17: This and above comment are stale.  There is no view.

src/Components/ReloadableComponent/ReloadableConnection.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+  public render() {
+    let ui: React.ReactNode;
+
+    if (!this.state.imodel)
+      ui = <StartupComponent iModelName={this.props.iModelName} onIModelReady={this._onIModelReady} />;
+
+    return (
+      <>
+        {ui}
```
Josh-Schifter, line 35: Let's talk about this.  It's inconsistent with the reloadable viewport.

Josh-Schifter:
I think we can do better on the instructions for these to describe what they do.  Also, let's consider changing the order.  For example, it seems like 'Presentation Tree' is the simplest of the new group so it should probably come first.

tomdicarlo:
I updated the way the scrollbar looks to be consistent with the tree scrollbar. It also matches much better with the dark theme of the components and sample gallery.

PR #32, opened 2020-06-26
Some samples default to different iModels

Josh-Schifter:
Some samples just look or work better with retail building or baytown.  We want the thumbnails to match with the starting model.
Also, I fixed a small bug with initializing the modelselector.


Josh-Schifter:
> Just out of interest, what was the bug caused by the other way of getting the iModelSelector?

When a new sample is selected, the iModelSelector should be initialized to the first iModel in the list.  Problem is that we were using the iModel list from the state which reflects the previous sample.  Need to use the list from the new sample that hasn't been pushed to state yet.

PR #36, opened 2020-06-29
Make the sample code easier to consume by splitting ui code into a separate file

Josh-Schifter:


Josh-Schifter:
I'm thinking SampleNameApp.tsx and SampleNameUI.tsx.  That matches the names of the classes that already exist, ex. MarkerPinsApp and MarkerPinsUI.  There is a class called HeatmapDecoratorUI_Component_ that I would rename to just HeatmapDecoratorUI.  All those UI classes have a method called getControlPane, but the classes also have (for example) the reference to the ReloadableViewport plus the onIModelReady and getCustomViewState.

Josh-Schifter:
In case it wasn't obvious I'm proposing files such as:
HeatmapDecoratorApp.tsx
HeatmapDecoratorUI.tsx
MarkerPinsApp.tsx
MarkerPinsUI.tsx

tomdicarlo:
I like the file naming convention. And are you saying we should keep getControlPane the same, since the UI file contains more than just the control pane?  If so, that makes sense to me, since the UI references the other elements you listed.

Josh-Schifter:
> And are you saying we should keep getControlPane the same, since the UI file contains more than just the control pane? If so, that makes sense to me, 

Exactly, yes.  I'll push that change to marker pins and heatmap.  Then we can look it over once more before replicating it everywhere else.

Josh-Schifter:
One more item for this branch... we should use the .ts extension for all files that don't use jsx.  For now we should prioritize the frontend-samples folder.

tomdicarlo:
Two notes on the changes I made:
1) On three samples (ViewerOnly2d, ViewClip, and ZoomToElement) all the code made sense to remain with the UI. I still created an application file, which is just a wrapper to the UI. I did it this way so all the sample specs would be standardized, with each of them calling the .setup method on the app file. Not sure if there are any thoughts on this, as I could certainly take a different approach.
2) I tried to keep the changes Stefan made as in tact as possible. That wasn't entirely possible since many files were broken up and renamed. Most of the changes that did occur are either class names or the paths to imports, but if that would somehow impact the editor, let me know.

src/frontend-samples/viewer-only-2d-sample/sampleSpec.ts
```diff
@@ -3,23 +3,20 @@
...
 
 export function getViewerOnly2dSpec(): SampleSpec {
   return ({
     name: "viewer-only-2d-sample",
     label: "2d",
     image: "viewer-only-2d-thumbnail.png",
     files: [
-      { name: "ViewerOnly2dSample.tsx", import: import("!!raw-loader!./index"), entry: true },
+      { name: "ViewerOnly2dApp.tsx", import: import("!!raw-loader!./ViewerOnly2dApp"), entry: true },
+      { name: "ViewerOnly2dUI.tsx", import: import("!!raw-loader!./ViewerOnly2dUI"), entry: true },
```
StefanRetief, line 16: I have concerns about the multiple entries when we introduce the transpiler, but that's a problem for the next PR. I think that having multiple entries will lead to having multiple bundles instead of  just one (much like how webpack handles it).

src/frontend-samples/view-clip-sample/ViewClipApp.tsx
```diff
@@ -0,0 +1,96 @@
...
+import { ClipMaskXYZRangePlanes, ClipPlane, ClipPrimitive, ClipShape, ClipVector, ConvexClipPlaneSet, Plane3dByOriginAndUnitNormal, Point3d, Vector3d } from "@bentley/geometry-core";
+
+export default class ViewClipApp {
+  public static async setup(iModelName: string, iModelSelector: React.ReactNode) {
+    return <ViewClipUI iModelName={iModelName} iModelSelector={iModelSelector} />;
+  }
+}
+
+/* Method for clearing all clips in the viewport */
+export function clearClips(vp: ScreenViewport) {
```
Josh-Schifter, line 19: Why export a bunch of separate functions?  The pattern from the other samples would be to put all of these into the App class which is already exported.

src/frontend-samples/zoom-to-elements-sample/ZoomToElementsApp.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+    // Set the view to point at a volume containing the list of elements
+    await vp.zoomToElements(elementIds, { ...viewChangeOpts, ...zoomToOpts });
+
+    // Select the elements.  This is not necessary, but it makes them easier to see.
+    imodel.selectionSet.replace(elementIds);
+  }
+}
+
+export default class ViewClipApp {
```
Josh-Schifter, line 22: ViewClipApp is wrong here.  Should be ZoomToElementsApp?

src/frontend-samples/zoom-to-elements-sample/ZoomToElementsApp.tsx
```diff
@@ -0,0 +1,44 @@
...
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
+import "../../common/samples-common.scss";
+import { IModelApp, IModelConnection, MarginPercent, ViewChangeOptions, Viewport, ZoomToOptions } from "@bentley/imodeljs-frontend";
+import ZoomToElementsUI, { ZoomToState } from "./ZoomToElementsUI";
+
+class ZoomToElementsAPI {
```
Josh-Schifter, line 11: This API class seems pointless.  I suggest lumping the one method into the APP class.

src/frontend-samples/thematic-display-sample/ThematicDisplayApp.tsx
```diff
@@ -0,0 +1,95 @@
...
+import * as React from "react";
+import { Range1dProps } from "@bentley/geometry-core";
+import { ThematicDisplay, ThematicDisplayProps, ThematicGradientColorScheme } from "@bentley/imodeljs-common";
+import { Viewport, ViewState3d } from "@bentley/imodeljs-frontend";
+import ThematicDisplaySampleUI from "./ThematicDisplayUI";
+
+// cSpell:ignore imodels
+
+/** This file contains the user interface and main logic that is specific to this sample. */
+export class ThematicDisplayAPI {
```
Josh-Schifter, line 15: This split between API and APP is interesting but we don't do it in the rest of the samples.  Seems better to follow the common pattern.

PR #37, opened 2020-06-29
Changed module export for samples to export default

StefanRetief:
Pre-Monaco-Editor upgrade for code editing and running. Code editor requires the modules that will be run to be default exports. All samples are now exported as defaults.

### DIFF
--- a/src/frontend-samples/component-gallery/badge-sample/index.tsx
+++ b/src/frontend-samples/component-gallery/badge-sample/index.tsx
@@ -13,7 +13,7 @@ import { BetaBadge, NewBadge } from "@bentley/ui-core";
 export const createComponentExample = (title: string, description: string | undefined, content: React.ReactNode): ComponentExampleProps => {
   return { title, description, content };
 };
-export class BadgeList extends React.Component<{}> {
+export default class BadgeList extends React.Component<{}> {
 
   // Combines several instances of ComponentExampleProps to be passed into the ComponentContainer
   public static getBadgeData(): ComponentExampleProps[] {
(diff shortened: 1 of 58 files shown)


Josh-Schifter:
Seems like you've made one class from each sample to be 'export default'.  Why that particular class?  What's the actual requirement here?

I'm asking because we are considering to reorganize all the view related samples to split the 'App' class from the 'UI' class.  When that's done, will we need to have two 'export default' classes?

The other reason I'm asking is because I was told that imodel.js in general tends to avoid export default.  So it'd be better if we didn't do it in the sample code.

See this PR for WIP reorg: https://github.com/imodeljs/frontend-sample-showcase/pull/36

StefanRetief:
> Seems like you've made one class from each sample to be 'export default'.  Why that particular class?  

The particular class was chosen as it’s the component that is returned by the setup method of the sample spec (I.e. the “ViewportOnlyUI” component is returned from the setup method of the getViewportOnlyUI sampleSpec)

>What's the actual requirement here?

When lazy loading react components, there isn’t a way to load the component as a named import just yet, it has to imported as default. This is necessary for running the code as after the component is transpiled and bundled, it is mounted to the dom by lazy loading in react. The only other method we could consider (but I personally oppose as it will make things messy) is to make an intermediate class that exports the named import as default.

> I'm asking because we are considering to reorganize all the view related samples to split the 'App' class from the 'UI' class.  When that's done, will we need to have two 'export default' classes?

We will only need the component that is returned by the setup method to be exported as default. I don’t think the split should be affected as we can load the “App” class as a module, like we do the ReloadableViewport component

Josh-Schifter:
So after the refactor, it will always be in the UI file.  I suppose that's not too bad since the App file is really the main one.

PR #43, opened 2020-07-03
Monaco editor upgrade

StefanRetief:
This PR updates the Monaco-Editor to enable transpiling and running the code from the code editor.


src/Components/SampleShowcase/SampleShowcase.tsx
```diff
@@ -141,36 +160,91 @@ export class SampleShowcase extends React.Component<{}, ShowcaseState> {
...
+    const sampleUi = await import( /* webpackIgnore: true */ blob);
+
+    if (!this._prevSampleSetup) {
+      this._prevSampleSetup = activeSample.setup;
+    }
+    if (!this._prevSampleTeardown) {
+      this._prevSampleTeardown = activeSample.teardown;
+    }
+
+    /* HACK: Because some samples have a setup method stored in the sample itself, which is editable, we need to make suree call it.
```
Josh-Schifter, line 182: We could refactor the samples so that the setup / teardown methods always live in the SampleSpec.tsx.  For some of them, those methods might need to call a method in the SampleApp class (because the setup and teardown logic is important for the user to see).  Would that help?

StefanRetief:
> I'm concerned about the long term difficulty of maintaining the .tsx files manually.  It's ok in the short term, but wouldn't it be better if those were generated at build time?

Which tsx files are you referring to?

src/Components/SampleShowcase/SampleShowcase.tsx
StefanRetief: I don’t think that would help much in this case, unless we also load the “App” portion of the sample as a “module” by referencing it in the modules.json. This would make the “App” portion of the sample uneditable, which I think starts taking away from the experience. What might work instead is if we created an interface and implemented it for each of the “App” portions of the sample. Then when it’s time to import, we can cast the default import as that interface. That would solve the need to search for the right method.

Probably something I should’ve done instead of this.

Josh-Schifter:
> > I'm concerned about the long term difficulty of maintaining the .tsx files manually.  It's ok in the short term, but wouldn't it be better if those were generated at build time?
> 
> Which tsx files are you referring to?

My mistake.  I meant .d.ts

src/Components/SampleShowcase/SampleShowcase.tsx
Josh-Schifter: >>if we created an interface and implemented it for each of the “App” portions of the sample

That is quite reasonable.

StefanRetief:
> > > I'm concerned about the long term difficulty of maintaining the .tsx files manually.  It's ok in the short term, but wouldn't it be better if those were generated at build time?
> 
> > 
> 
> > Which tsx files are you referring to?
> 
> 
> 
> My mistake.  I meant .d.ts

Ah, right. So there is a new script provided by the Monaco editor, that generates the .d.ts files based on the modules.json. I’ve changed the build and start scripts to include this script. Is this what you had in mind?

Josh-Schifter:
> Ah, right. So there is a new script provided by the Monaco editor, that generates the .d.ts files based on the modules.json. I’ve changed the build and start scripts to include this script. Is this what you had in mind?

Sounds right.  Do you mean you are planning to change the build?  if you already did it, then we shouldn't need to check them into the repo, right?

StefanRetief:
> Sounds right.  Do you mean you are planning to change the build?  if you already did it, then we shouldn't need to check them into the repo, right?

I already changed it, so I can remove the existing .d.ts files before I complete the merge and add them to the .gitignore.

src/Components/SampleShowcase/SampleShowcase.tsx
StefanRetief: I've updated to use a "SampleApp" abstract class that each sample implements. I'm able to cast the import to this type and it works nicely to allow the showcase to call the correct method.

PR #44, opened 2020-07-06
Fix sheet border size

pmconne:
ViewCreator2d was mistakenly assuming that the sheet size is the same as the sheet model's range. The model range is the union of all of its elements ranges. The sheet size is typically larger than that so that it can encompass the contents with a margin (but can also be smaller because sometimes people draw outside the border).

Fix: Query the iModel for the actual sheet size instead.


roopsaini:
I did…but wasn’t merged in since porting samples to showcase!

Thanks @Paul Connelly<mailto:Paul.Connelly@bentley.com>

PR #49, opened 2020-07-09
Fixed 2dViewer sample to start on a 2d model

tomdicarlo:


src/frontend-samples/viewer-only-2d-sample/ViewerOnly2dApp.tsx
```diff
@@ -10,19 +10,38 @@ import { IModelApp, IModelConnection } from "@bentley/imodeljs-frontend";
...
 import { ModelProps } from "@bentley/imodeljs-common";
 import SampleApp from "common/SampleApp";
+import { ViewSetup } from "api/viewSetup";
 
 export default class ViewerOnly2dApp implements SampleApp {
   public static async setup(iModelName: string, iModelSelector: React.ReactNode) {
     return <ViewerOnly2dUI iModelName={iModelName} iModelSelector={iModelSelector} />;
   }
+
+  public static getDrawingModelList(models: ModelProps[]) {
```
Josh-Schifter, line 20: Moving this to app is supposed to illustrate how a user would query for the list of models.  As it is, it just takes the list that is passed in and turns it into UI objects.

Turning into UI objects belongs in the UI.tsx.  Querying for the list belongs in the App.tsx.  It's backwards now.

PR #57, opened 2020-07-22
Panel collapse

tomdicarlo:
Enables the collapsing and expanding of the control pane and the sample gallery.

This required an overhaul of the control pane, which now is constructed in SampleShowcase.tsx, where it receives the sample's instructions and controls through a callback function.


BaoTon:
This slider looks great!

Is it possible to add an "exit" button to the gallery panel to force an instant collapse? 
I think most users would prefer a button to instantly close the panel rather than being forced to drag to close the panel.

BaoTon:
i.e. our editor does the sliding collapse, as well as providing an "X" toggle to close it immediately.

tomdicarlo:
Yes that is definitely something that would be helpful. The only worry I have about there would be where on the gallery we would put the button, but I'll experiment with a few designs to find something that looks good.

If you don't have any other issues with the current PR, could I add that button in a separate PR? I won't be here tomorrow, and I don't think I'll be able to finish that by the end of the day. So I was hoping to get this through before I leave

PR #63, opened 2020-07-30
Add tests

williamkbentley:
Adds a basic set of tests using certa.


README.md
```diff
@@ -37,6 +37,12 @@ Instead, it will copy all the configuration files and the transitive dependencie
 
 You don’t have to ever use `eject`. The curated feature set is suitable for small and middle deployments, and you shouldn’t feel obligated to use this feature. However we understand that this tool wouldn’t be useful if you couldn’t customize it when you are ready for it.
 
+## Testing
+
+This package uses a Bentley developed testing tool [Certa](https://www.npmjs.com/package/@bentley/certa). The three required scripts to run the tests are `test-build`, `webpackTests`, and finally `test-certa`. create-react-app does not contain the necessary framework for testing an iModel.js application which is the reason for the test specific build and webpack workflow.
```
BaoTon, line 42: Could you further elaborate on what each script does?

williamkbentley:
> Overall a nice setup.
> 
> A couple of things:
> 
> 1. Do you think splitting up the frontend-sample-showcase.test.tsx makes sense? This file will get larger and I imagine separating the contexts will help better organize things.
> 2. If we split up the file - can we run targeted tests? I think a webpack of the entire test suit occurs for every change correct? This may contribute to poor test times as the project grows - I'm not sure.
> 3. This might annoy everyone but maybe it's a good idea to force at least 1 test per sample to get the build passing? Will probably improve the integrity of the code. Probably a better idea is a WARNING.
> 
> Thanks!

williamkbentley:
1. Do you think splitting up the frontend-sample-showcase.test.tsx makes sense? This file will get larger and I imagine separating the contexts will help better organize things. **_WB: This PR was just intended to get the framework and provide some examples. As the list grows, we can adjust the file structure._**
1. If we split up the file - can we run targeted tests? I think a webpack of the entire test suit occurs for every change correct? This may contribute to poor test times as the project grows - I'm not sure. _**WB: Unfortunately yes, the entire source and tests have to be re-webpacked each time. However, you are able to use a [variety of mocha options](https://github.com/imodeljs/imodeljs/blob/master/tools/certa/bin/certa.js#L13) with certa. Such as grep or fgrep.**_
1. This might annoy everyone but maybe it's a good idea to force at least 1 test per sample to get the build passing? Will probably improve the integrity of the code. Probably a better idea is a WARNING. _**WB: I don't know how we can force this programmatically.  In addition, tests like these do not make sense for all samples - frontend-only and some of the UI components for example. These tests should be used when the sample uses the iModel.js library to verify that the Viewport or some other result has been updated.**_

PR #68, opened 2020-08-05
Add header

BaoTon:
Add header to the sample-showcase.

These changes will only be apparent on deploy. As such, NugetPackage.nuspec will now bundle files into /sample-showcase/frontend-sample-showcase, and /sample-showcase/index.html now links to an iframe.

There is an added change in window.history.replaceState to accommodate for URL parameters propagating up to the parent page.


BaoTon:
Do not approve yet - build pipeline in Azure needs to be changed.

public/iframe-index.html
```diff
@@ -0,0 +1,122 @@
...
+                    
+                    <div class="navbar-collapse collapse justify-content-stretch" id="navbar5">
+                        <ul class="navbar-nav mb-0" id="doc-site-tabs">
+                                <li class="nav-item"><a class="nav-link" href="/getting-started">Getting Started</a></li><li class="nav-item active"><a class="nav-link" href="/sample-showcase">Samples</a></li><li class="nav-item "><a class="nav-link" href="/learning">Learning</a></li><li class="nav-item "><a class="nav-link" href="/reference">API</a></li><li class="nav-item dropdown change-history-tab  "><a class="nav-link" style="display:inline-block;" href="/changehistory">v2.4.0</a><a class="nav-link dropdown-toggle" style="display:inline-block;padding-left:1px;margin-left:1px;box-shadow:none;" href="/changehistory" id="navbarDropdown" role="button" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false"></a>
+                               <div class="dropdown-menu" aria-labelledby="navbarDropdown"><a class="dropdown-item" href="/changehistory">v2.4.0</a><a class="dropdown-item" href="/v1/getting-started">v1.14.0</a><div class="dropdown-divider"></div>
```
williamkbentley, line 95: Are we going to have to update this with each release?

BaoTon: Yes, that "2.4.0" part. 

I think it's a good idea to write a dummy header page with "REPLACE_ME" every time we build. Or simply write a script that replaces the version with the current frontend version pulled from package.json. I'm not sure the best approach yet.

BaoTon: It probably makes sense to pull in a new bootstrap page and not have this maintained in this repo. The only code I need to add is in the main-wrapper:

        <div id='main-wrapper' style="height: 1px; display: block">
          <iframe id="frontend-samples" width=100% height= 100% src=""></iframe>
        </div>

      <script>
        function receiveMessage(event) {
          window.history.replaceState(null, "", event.data);
        }
        window.addEventListener("message", receiveMessage, false);

        document.getElementById("frontend-samples").src = "/sample-showcase/frontend-sample-showcase/" + window.location.search;
      </script>

BaoTon: Then again, this is probably the only unique page we'll have on the entire docs. I'm not sure writing a template page is worth it yet.

williamkbentley: I'm not sure what you mean. Making a BeMetalsmith template? Also, pulling from package.json is probably not a good idea. The released iModel.js version can be different than the version the sample-showcase is using (although it probably shouldn't be).

BaoTon: Not template, a new layout* sorry. Refer to my other PR.

PR #81, opened 2020-08-14
Color and icon updates

Josh-Schifter:


Josh-Schifter:
> Is the eye icon still supposed to be used for the control pane? I think it looks good, but wasn't sure if you meant to change that. Otherwise everything looks good to me!

No that was a merging error.  Fixed now.

PR #85, opened 2020-08-19
Reality modeling

williamkbentley:
Sample for displaying reality modeling. Shows how to attach reality models to an iModel.js app. Currently only has one sample iModel - Exton campus.


Josh-Schifter:
Initial painting of the reality data seems extremely slow to me.  Several minutes before I can see all the trees around the Exton Campus.  Might be my network.

PR #87, opened 2020-08-25
Update to 2.5.0 - Fix for NoSigninAuthClient missing _tokenString with new IMS 2.0 update.

BaoTon:
- NoSigninAuthClient missing _tokenString key
- Patched ColorPicker to use initialColor instead of ActiveColor.
- Add @bentley/rbac-client as a dependency


BaoTon (commented):
Awaiting 2.5.2 update to fix a known tile caching issue.

PR #88, opened 2020-08-26
Display Style Sample

Ellord207:


BaoTon:
> I'm getting the error: Property 'overrideDisplayStyle' does not exist on type 'Viewport'. Looks like that method was added in 2.3. Bao is going to update the showcase to 2.5.3 soon so maybe just wait for that.

2.5.3 has already been merged to master - you'll need to update the branch.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
```diff
@@ -0,0 +1,106 @@
...
+      DisplayStylesApp.applyDisplayStyle(this.state.viewport, style);
+      if (this.state.merge && CUSTOM_STYLE_INDEX !== this.state.activePresetIndex) {
+        const custom = this.props.displayStyles[CUSTOM_STYLE_INDEX];
+        DisplayStylesApp.applyDisplayStyle(this.state.viewport, custom);
+      }
+    }
+  }
+
+  private getControls(): React.ReactNode {
+    const toggleTooltip = "Toggling on will apply the \"Custom\" style in \"Styles.ts\" after apply the selected style.";
```
Josh-Schifter, line 75: Grammar:
Toggling on will apply the \"Custom\" style in \"Styles.ts\" after the selected style is applied.

src/frontend-samples/display-styles-sample/DisplayStylesApp.tsx
```diff
@@ -0,0 +1,33 @@
...
+import DisplayStylesUI from "./DisplayStylesUI";
+import { displayStyles } from "./Styles";
+
+/** This sample highlights the ability to override the current display settings with a JObject or
+ * JSON describing the desired settings. All the descriptions are in the Styles.ts file.
+ *
+ * We invite you to create your own settings using the "Custom" tag provided in Styles.ts.
+ * You can copy and paste parts of your favorite styles as a start.
+ *
+ * Only the features specificized by the style will be changed. This can be seen with the "Custom"
```
Josh-Schifter, line 20: Spelling: specified

src/frontend-samples/display-styles-sample/DisplayStylesApp.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * style you can create. Use the "Merge with Custom" toggle to apply both the custom and the specified
+ * styles.
```
Josh-Schifter, line 22: Grammar:
Use the "Merge with Custom" toggle to apply both the custom and the specified style.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  // Called by the control and updates wether to also apply the Custom display style.
+  private readonly _onToggle = (isOn: boolean) => {
+    this.setState({ merge: isOn });
+  }
+
+  // Will be triggered once when the iModel is loaded.
+  private readonly _onIModelReady = (_iModel: IModelConnection) => {
+    const vp = IModelApp.viewManager.selectedView;
+    if (undefined === vp)
```
Josh-Schifter, line 46: Is this test really needed?  Other samples just always use view passed to onViewOpen.  Ex. Heatmap

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
```diff
@@ -0,0 +1,106 @@
...
+  public componentDidUpdate(_prevProp: DisplayStylesUIProps, prevState: DisplayStylesUIState) {
+    let updateDisplayStyle = false;
+    if (this.state.viewport?.viewportId !== prevState.viewport?.viewportId)
+      updateDisplayStyle = true;
+    if (this.state.activePresetIndex !== prevState.activePresetIndex)
+      updateDisplayStyle = true;
+    if (this.state.merge !== prevState.merge)
+      updateDisplayStyle = true;
+
+    if (updateDisplayStyle && undefined !== this.state.viewport) {
```
Josh-Schifter, line 64: testing for undefined !== viewport is questionable since an undefined viewport will crash the program a few lines up.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
```diff
@@ -0,0 +1,106 @@
...
+        this.setState({ viewport });
+      });
+    else
+      this.setState({ viewport: vp });
+  }
+
+  /** A render method called when the state or props are changed. */
+  public componentDidUpdate(_prevProp: DisplayStylesUIProps, prevState: DisplayStylesUIState) {
+    let updateDisplayStyle = false;
+    if (this.state.viewport?.viewportId !== prevState.viewport?.viewportId)
```
Josh-Schifter, line 57: I suggest testing for undefined viewport and quick exit at the top of this method.  This line will crash on undefined viewport.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
```diff
@@ -0,0 +1,106 @@
...
+        const custom = this.props.displayStyles[CUSTOM_STYLE_INDEX];
+        DisplayStylesApp.applyDisplayStyle(this.state.viewport, custom);
+      }
+    }
+  }
+
+  private getControls(): React.ReactNode {
+    const toggleTooltip = "Toggling on will apply the \"Custom\" style in \"Styles.ts\" after apply the selected style.";
+    const entries = this.props.displayStyles
+      .map((styles) => styles.name)
```
Josh-Schifter, line 77: Why two calls to map?  This can be done in one call.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
```diff
@@ -0,0 +1,106 @@
...
+          <span>Merge with Custom:</span>
+        </span>
+        <Toggle isOn={this.state.merge} onChange={this._onToggle} />
+      </div>
+    );
+  }
+
+  /** The sample's render method */
+  public render() {
+    const instruction = "Use the drop down below to change the display styles. Edit the \"Custom\" style in \"Style.ts\" and re-run the sample to see the changes.";
```
Josh-Schifter, line 96: Shouldn't it be 'display style'  (singular)?

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
Ellord207: The ["?." operator (Optional Chaining)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) will return undefined if the viewport is undefined.  This case handles both a change to the viewport (2 =/= 1) and initialization of the viewport (1 =/= undefined).  Can confirm with chrome debugger.
I did it this way so the controls can be live while the viewport is loading, then the current state of the controls are applied to the viewport when finished loading.  The intent is to minimize downtime if the user is trying create a custom style in styles.ts and is reloading after running updated code.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
Ellord207: Does not crash from a few lines up.  See reply about the issue there.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
Josh-Schifter: I stand corrected.

src/frontend-samples/display-styles-sample/DisplayStylesUI.tsx
Ellord207: Probably not.  Without it, it does (theoretically) create a race condition.  Will remove for consistency.

PR #91, opened 2020-09-07
Created volume query sample

EimantasDaknys:


src/frontend-samples/query-volume-app/VolumeQueryUI.tsx
```diff
@@ -267,10 +267,10 @@ export default class VolumeQueryUI extends React.Component<
         </div>
         <div className="table-wrapper">
           <select multiple style={{ maxHeight: "100px", overflowY: "scroll", overflowX: "hidden", whiteSpace: "nowrap" }}>
-            {Array.from(this.state.elementsToShow.elements).map((element) => <option key={element.id} style={{ listStyleType: "none", textAlign: "left" }}>{element.name}</option>)}
+            {this.state.elementsToShow.elements.slice(99).map((element) => <option key={element.id} style={{ listStyleType: "none", textAlign: "left" }}>{element.name}</option>)}
```
Josh-Schifter, line 270: This is omitting the first 100 elements.  Should be slice(0,99).

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
```diff
@@ -0,0 +1,179 @@
...
+    const packsOfInstanceKeys = Math.floor(instanceKeys.length / 1000);
+    for (let i = 0; i <= packsOfInstanceKeys; i++) {
+      if (i !== packsOfInstanceKeys) {
+        packedInstancesKeys.push(instanceKeys.slice(i * 1000, (i + 1) * 1000));
+      } else {
+        packedInstancesKeys.push(instanceKeys.slice(i * 1000, instanceKeys.length + 1));
+      }
+    }
+
+    /* Getting elements names */
```
Josh-Schifter, line 116: Maybe worth a comment here to say, the reason why you created the packages was to avoid making a single query for a very large number of labels.

Also... why are you getting the labels for all the elements if you are only going to show the first 100?

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    for (let i = 0; i < elementNames.length; i++)
+      elements[i].name = elementNames[i];
+
+    return elements;
+  }
+
+  /* Clear color from colored elements */
+  public static clearColoredSelection(vp: ScreenViewport) {
```
Josh-Schifter, line 130: Please rename to clearColorOverrides.  This really has nothing to do with Selection.

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
```diff
@@ -0,0 +1,179 @@
...
+    EmphasizeElements.getOrCreate(vp).clearOverriddenElements(vp);
+  }
+
+  /* Sort given elements - inside, outside the box and overlap */
+  public static async getSortedSpatialElements(vp: ScreenViewport, candidates: PhysicalElement[]): Promise<SpatialElements | undefined> {
```
Josh-Schifter, line 135: Please rename to classifyElements.  We typically use the term "sort" for setting the ordering of a set.  Ex. 1,3,2 => 1,2,3

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
```diff
@@ -0,0 +1,179 @@
...
+import { EmphasizeElements, IModelApp, ScreenViewport, ViewClipClearTool, ViewClipDecorationProvider, ViewClipTool } from "@bentley/imodeljs-frontend";
+import { ClipMaskXYZRangePlanes, ClipPlaneContainment, ClipShape, ClipUtilities, ClipVector, Range3d } from "@bentley/geometry-core";
+import SampleApp from "common/SampleApp";
+import VolumeQueryUI from "./VolumeQueryUI";
+import { ColorDef, GeometryContainmentRequestProps } from "@bentley/imodeljs-common";
+import { BentleyStatus, Id64Array } from "@bentley/bentleyjs-core";
+import { PresentationLabelsProvider } from "@bentley/presentation-components";
+import { InstanceKey } from "@bentley/presentation-common";
+
+export interface SpatialElements {
```
Josh-Schifter, line 16: Please rename to classifiedElements.  SpatialElement is a BIS class.

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
```diff
@@ -0,0 +1,179 @@
...
+  insideTheBox: PhysicalElement[];
+  outsideTheBox: PhysicalElement[];
+  overlap: PhysicalElement[];
+}
+
+export interface SpatialElementsColors {
```
Josh-Schifter, line 22: Spatial could be problematic here too.  Maybe better to call it ClassificationColors.

src/frontend-samples/query-volume-app/VolumeQueryUI.tsx
```diff
@@ -0,0 +1,287 @@
...
+    const range = VolumeQueryApp.getRangeOfBox(vp);
+    VolumeQueryApp.clearClips(vp);
+    VolumeQueryApp.addBoxRange(vp, range, isToggleOn);
+    /* If Volume box is off - turning it on */
+    if (isToggleOn && !this.state.isVolumeBoxOn)
+      this.setState({ isVolumeBoxOn: true });
+  }
+
+  /* Coloring elements that are inside, outside the box or overlaping */
+  private _onClickApplyColoring = async () => {
```
Josh-Schifter, line 88: Please rename _onClickApplyColorOverrides

src/frontend-samples/query-volume-app/VolumeQueryUI.tsx
```diff
@@ -0,0 +1,287 @@
...
+
+    /* Getting elements that are going to be colored */
+    const candidates = await VolumeQueryApp.getAllSpacialElements(vp);
+    const spatialElements = await VolumeQueryApp.getSortedSpatialElements(vp, candidates) as SpatialElements;
+    await VolumeQueryApp.colorSpatialElements(vp, spatialElements, this.state.spatialElementsColors);
+    this.setState({ spatialElements });
+    await this._setListOfElements(vp);
+  }
+
+  private _onClickClearColoring = () => {
```
Josh-Schifter, line 105: please rename _onClickClearColorOverrides

src/frontend-samples/query-volume-app/VolumeQueryUI.tsx
```diff
@@ -0,0 +1,287 @@
...
+    if (undefined === vp) {
+      return false;
+    }
+
+    VolumeQueryApp.clearColoredSelection(vp);
+    /* Emptying elements to show list*/
+    this._emptyListOfElements();
+  }
+
+  /* Changing colors of elements that are going to be selected */
```
Josh-Schifter, line 116: The word selected is wrong here.   Selected refers to the user selecting elements, ex. using the element selection tool.  This is changing the colors that will be used for overrides.

src/frontend-samples/query-volume-app/VolumeQueryUI.tsx
```diff
@@ -0,0 +1,287 @@
...
+  private _onIModelReady = (imodel: IModelConnection) => {
+    this.setState({ imodel });
+
+    IModelApp.viewManager.onViewOpen.addOnce((_vp: ScreenViewport) => {
+      this.setState({ imodel, isVolumeBoxOn: true }, () => { this._onToggleVolumeBox(true); });
+      // tslint:disable-next-line no-floating-promises
+      this._onClickApplyColoring();
+    });
+  }
+  /* Setting elements that are going to be showed, based of given position*/
```
Josh-Schifter, line 163: Add one empty line between methods.

src/frontend-samples/query-volume-app/sampleSpec.tsx
```diff
@@ -0,0 +1,25 @@
...
+    files: [
+      { name: "VolumeQueryApp.tsx", import: import("!!raw-loader!./VolumeQueryApp"), entry: true },
+      { name: "VolumeQueryUI.tsx", import: import("!!raw-loader!./VolumeQueryUI") },
+    ],
+    customModelList: [
+      SampleIModels.RetailBuilding,
+      SampleIModels.BayTown,
+      SampleIModels.House,
+    ],
+    setup: VolumeQueryApp.setup,
```
Josh-Schifter, line 23: needs to be VolumeQueryApp.setup.bind(VolumeQueryApp)

Josh-Schifter (changes requested):
I left comments mainly related to naming.  You are querying for PhysicalElements and you are using 'Spatial' to mean classified based on inside/outside/overlap, but SpatialElement is a BIS class.

Probably you should be querying for SpatialElements.  I suggest the term ClassifiedElements for the grouping based on inside/outside/overlap.

src/frontend-samples/query-volume-app/VolumeQueryUI.tsx
```diff
@@ -0,0 +1,287 @@
...
+            <option value={ElementPosition.OutsideTheBox}> {ElementPosition.OutsideTheBox} </option>
+            <option value={ElementPosition.Overlap}> {ElementPosition.Overlap} </option>
+          </select>
+        </div>
+        <div className="table-wrapper">
+          <select multiple style={{ maxHeight: "100px", overflowY: "scroll", overflowX: "hidden", whiteSpace: "nowrap" }}>
+            {this.state.elementsToShow.elements.slice(99).map((element) => <option key={element.id} style={{ listStyleType: "none", textAlign: "left" }}>{element.name}</option>)}
+          </select>
+        </div>
+        <span style={{ color: "grey" }} className="table-caption">Selected {this.state.elementsToShow.elements.length} elements, showing top 100.</span>
```
Josh-Schifter, line 273: These elements are not selected.  Please change the message to:  "List contains <count> elements, showing first 100."

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
EimantasDaknys: Well I am getting all the labels because I have interface PhysicalElement, that has id, name, className as its properties. The program works like this(just changed it for performance) - when you select iModel, all physicalElements are loaded and put into VolumeQueryUI class property with ids, names and classNames. Then when you change the volume of the box and click apply, you don't load all the elements again. You take them from VolumeQueryUI property and pass them to API so it would show which ones are inside/outside/overlap. In other words, firstly you load all the PhysicalElements, and then you work with them(list them or do other operations). So getting only 100 elements names would not work in this scenario. Also the main reason why the slice is being done is not because we load too much elements and it takes time, but because representing 1000 elements in list takes ~1 sec(you can feel it when changing what elements you want to see - inside/outside/overlap).

src/frontend-samples/query-volume-app/VolumeQueryApp.tsx
```diff
@@ -103,7 +103,8 @@ export class VolumeQueryApp implements SampleApp {
       return { className: element.className.replace(".", ":"), id: element.id };
     });
 
-    /* Sorting elements into packages. Instead of one 3000 length array I will have 3 x 1000 lenght arrays */
+    /* Breaking one array of 3000 keys into 3 arrays with 1000 keys each*/
```
Josh-Schifter, line 106: This comment still implies that there are always 3000 keys.  But that is not true.  The number of physical elements in every iModel is different.  The comment should say something like: "Break up the potential large array into smaller arrays with a maximum of 1000 keys each.  For example, if there are 3000 physical elements, this will create 3 arrays with 1000 keys each."

Josh-Schifter (changes requested):

Josh-Schifter (approved):

BaoTon (approved):

williamkbentley (approved):

PR #94, opened 2020-09-21
Added loading bar

EimantasDaknys:


src/frontend-samples/volume-query-sample/VolumeQueryHelper.tsx
```diff
@@ -0,0 +1,93 @@
+/*---------------------------------------------------------------------------------------------
```
Josh-Schifter, line 1: I think this file should be named after the class it contains: ProgressBar.tsx.

Also, I think this file should be added to the SampleSpec so it is shown in the editor.

Josh-Schifter (changes requested):

Josh-Schifter (approved):

PR #97, opened 2020-09-23
Property formatting sample

Josh-Schifter:


grigasp:
Examples for approaches #1 and #3 look fine to me, but maybe I'd change the comments that explain when to use which approach:

- Approach #1 (using the `PresentationPropertyDataProvider`) should be used when the results are needed on the frontend and are going to be displayed in `PropertyGrid` component. The data provider returns properties of just one element - if properties for more than one element are requested - we merge the values. This approach does not prevent consumers from customizing the content - they can have a decorator provider which adds/modifies or removes properties.

- Approach #3 (requesting content straight from `PresentationManager`) should be used when results are needed on the backend (where we don't have `PropertyRecord` API) or when we want to display them in third-party components that don't use `PropertyRecord` objects (although in the latter case it might be easier to just use either `PresentationPropertyDataProvider` or `PresentationTableDataProvider` and map PropertyRecords to types supported by those other components).

I'm not sure about approach #2. It uses `ContentBuilder` which at this moment is `@internal`. And I'm not sure there are any advantages of this approach over using our data providers. 

Speaking of `@internal` APIs, `DEFAULT_PROPERTY_GRID_RULESET` also falls into that group and it's not even exported through the barrel file. I personally would like it to stay internal and ask consumers to write their own rulesets - that allows me to change our component's rules without affecting others. Also, it would make consumers think what they need when they have to come up with rules for their components. But maybe for the sake of simplicity this one should be available for them?... What do you think?

BTW, `PresentationTableDataProvider` is probably also worth being mentioned - it's similar to `PresentationPropertyDataProvider` in a way that it returns PropertyRecords, but it can return properties for multiple elements (not merged).

Josh-Schifter:
The original idea behind approach #2 was to point to the code in ContentBuilder, but necessary so they would call it, but so they could look through it and understand it’s logic.  That was the way I learned about Content and was able to create approach #3.  One idea is to remove the current #2 and instead just use the comments in approach #3 to refer to ContentBuilder.  I don’t love that because I really learned a lot when writing #2.  I’m open for other suggestions.  What do you think?

Do you think I should add a variant of approach #1 to show customization via decorator?  I didn’t know about that.  To add it, I’d appreciate some pointers.

I chose to avoid PresentationTableDataProvider only because it wouldn’t fit nicely in the UI.  Thinking now, that’s not really a good reason, so I’ll consider adding it in.

Regarding DEFAULT_PROPERTY_GRID_RULESET, it seems to me that we need some default ruleset.  We can’t ask everyone to craft their own.  The question that lead to this sample was “how can I show, in my own UI, the same properties as Design Review”.

grigasp:
Regarding #1 customization with a decorator - I think the most difficult part is to figure out an example that makes sense. Decorating the provider as simple as overriding the `getData` method:
```ts
class MyProvider extends PresentationPropertyDataProvider {
  public async getData() {
    const data = await super.getData();
    // ...customize the data here...
    return data;
  }
}
```
Maybe a good example would be to put all (or some) records under some fake `PropertyCategory`.


Regarding #2 - the logic in `ContentBuilder` is specific to *Presentation types -> Bentley UI types* conversion. When converting *Presentation types* to other kind of UI types, the logic is likely to be different. It seems to me that example #3 is enough to show the basics of that process, and Presentation team should improve documentation of its types, explaining what stuff means and possibly giving some small examples similar to what's used in `ContentBuilder`.


I removed the `@internal` flag from `DEFAULT_PROPERTY_GRID_RULESET` and exported it through the barrel file. The change is going to be available in 2.7 release.

Josh-Schifter:
Please take another look.  I made two significant changes:
- Updated approach 1 with the custom data provider overriding getData method
- Removed the ContentBuilder approach and replaced it with PresentationTableDataProvider

I know of two issues:
1. The property grid in approach one does not update when the Customize toggle is changed.
2. The table in approach two appears to be empty until you scroll horizontally.  It's not a bug, there's just many empty columns.

I would appreciate suggestions on both of those issues.  Thanks.

grigasp:
1. That's a bug in PropertyGrid component. I opened a PR for a fix, but actually we're going to deprecate the `PropertyGrid` component soon in favor of `VirtualizedPropertyGrid` which can handle larg numbers of properties much better. There's also a `VirtualizedPropertyGridWithDataProvider` component whose props are the same as of `PropertyGrid`, so I would suggest switching to that rather than waiting for a release with my fix.

2. This seems more like a UX question.. Maybe there should be some cue telling users that there are properties on the right?..

src/frontend-samples/property-formatting-sample/readme.md
```diff
@@ -0,0 +1,30 @@
...
+
+This sample demonstrates how to query and present element property data suitable for end users.
+
+## Purpose
+
+The purpose of this sample is to demonstrate the following:
+
+- Using the [PropertyGrid](https://www.itwinjs.org/reference/ui-components/propertygrid/propertygrid/) react component along with [PresentationPropertyDataProvider](https://www.itwinjs.org/reference/presentation-components/propertygrid/presentationpropertydataprovider/) to present the default set of properties for an iModel element.
+- Creating a subclass of PresentationPropertyDataProvider to customize the properties that are presented in the property grid.
+- Using the [Table](https://www.itwinjs.org/reference/ui-components/table/table/) react component together with [PresentationTableDataProvider](https://www.itwinjs.org/reference/presentation-components/table/presentationtabledataprovider/) to present a table containing the the default set of properties for a set of iModel elements.
```
williamkbentley, line 13: "the the"

src/frontend-samples/property-formatting-sample/readme.md
```diff
@@ -0,0 +1,30 @@
...
+- Using [Presentation.getContent](https://www.itwinjs.org/reference/presentation-frontend/core/presentationmanager/getcontent/) to directly query the formatted property data so that you can use it independently of our user interface components.
+
+## Description
+
+Even though you can access all of an iModel's data via ECSQL queries, the properties returned that way are not directly suitable for presentation to an end user.  For one thing, individual properties must typically be formatted specifically for presentation purposes (ex. unit conversions and date-time conventions).  Also, the user's perception of a single element doesn't always match the raw data storage (ex. [relationships](https://www.itwinjs.org/bis/intro/relationship-fundamentals/) and [element aspects](https://www.itwinjs.org/bis/intro/elementaspect-fundamentals/)).  For these reasons, it is often necessary to use the [Presentation Library](https://www.itwinjs.org/learning/presentation/).
+
+This sample illustrates three approaches for showing element properties.  Which approach to choose depends on you goal and what if any customization you need for your use case.
```
williamkbentley, line 20: "your goal"

src/frontend-samples/property-formatting-sample/PropertyFormattingApp.tsx
```diff
@@ -0,0 +1,157 @@
...
+  public static createTableDataProvider(keys: KeySet, imodel: IModelConnection) {
+    const dataProvider = new PresentationTableDataProvider({ imodel, ruleset: DEFAULT_PROPERTY_GRID_RULESET });
+    dataProvider.keys = keys;
+    return dataProvider;
+  }
+
+  /* Approach 3: Do it all yourself.  If you want to use your own UI components.
+
+  This approach shows how to process the Content object yourself.  This is an oversimplified implementation that
+  2does not handle the difficult cases like properties which are arrays or structs.  Instead it represents
```
williamkbentley, line 102: typo

williamkbentley (approved):
Some typos. But looks good to me

src/frontend-samples/property-formatting-sample/PropertyFormattingApp.tsx
```diff
@@ -0,0 +1,157 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
+
+import "../../common/samples-common.scss";
+import { Presentation, SelectionChangesListener } from "@bentley/presentation-frontend";
+import { DEFAULT_PROPERTY_GRID_RULESET, PresentationPropertyDataProvider } from "@bentley/presentation-components/lib/presentation-components/propertygrid/DataProvider"; // tslint:disable-line: no-direct-imports
```
grigasp, line 10: These are now exported through the barrel file

grigasp (changes requested):

grigasp (approved):

roopksaini (approved):

PR #102, opened 2020-10-06
Updated imodel.js version to 2.7.0

tomdicarlo:
Looked over each sample and did not find any issues caused by the new updates. I'll leave this open for a little while to give time for people to double check their own samples/code for issues caused by 2.7.0


Josh-Schifter (changes requested):
The app-ui samples aren't working.  I see a message that says "general.no-content" instead of the viewport.  Maybe @NancyMcCallB could comment.

Josh-Schifter (approved):

Josh-Schifter:
I deleted node-modules and started over.  The app-ui problem did not reproduce.

PR #110, opened 2020-10-20
Added reality data transparency slider

williamkbentley:


Josh-Schifter (approved):
Display of reality data is still very slow.  Transparency works well.

PR #112, opened 2020-10-22
Divided viewport

Ellord207:


Ellord207:
There is an bug I found where, if you lock the clip and change the comparison type, the comparison does not change.

pmconne (approved):
I have only nitpicks, which I will spare you. Good readable example.

williamkbentley (approved):

Josh-Schifter (changes requested):
The UI is inconsistent with the other samples.

- Looks like you are using the wrong selector component
- The spacing between the labels and the other components is wrong.is wrong.

src/frontend-samples/swiping-viewport/Divider.tsx
```diff
@@ -0,0 +1,158 @@
...
+      width = parseInt(widthStr, 10);
+    return width;
+  }
+  private limitToBounds(n: number): number {
+    n = Math.min(n, this.props.bounds.right - (this._elem!.clientWidth + this._buffer));
+    n = Math.max(n, this.props.bounds.left + this._buffer);
+    return n;
+  }
+
+  constructor(props: DividerComponentProps, context: any) {
```
Josh-Schifter, line 56: Remove the "context" argument.  The code editor in the showcase reports this as a "problem" at runtime.

src/frontend-samples/swiping-viewport/SwipingComparisonApp.tsx
```diff
@@ -0,0 +1,301 @@
...
+  GraphicBranch,
+  IModelApp,
+  IModelConnection,
+  RenderClipVolume,
+  SceneContext,
+  ScreenViewport,
+  TiledGraphicsProvider,
+  TileTreeReference,
+  Viewport,
+} from "@bentley/imodeljs-frontend";
```
Josh-Schifter, line 35: I think we should collapse each of these import statements into a single line.  At present, the imports take up an entire screen in the code editor at runtime.

Josh-Schifter (changes requested):

Josh-Schifter (approved):

PR #113, opened 2020-10-23
Fix editor reported issues

Josh-Schifter:
General cleanup including:
- Remove Modules.ts from the repo and add to .gitignore.  The file is generated at build time.
- Fix all unresolved imports that were being flagged by the embedded editor.
- Fix *almost* all 'problems' reported by the embedded editor.
- Make the compiler in tsconfig more strict so we catch more problems at build time.
- Clean up incorrect reference to readmes in app-ui samples.


StefanRetief (approved):

williamkbentley (approved):
LGTM

PR #115, opened 2020-11-06
Add Metrostation to reality-data-sample

williamkbentley:


Josh-Schifter (approved):
It looks good.  Performance is not great, but might be my bandwidth.

Maybe we should try a different default view?  Current one doesn't show the city.

PR #121, opened 2020-11-11
Geometry samples

tomdicarlo:
Visual Overhaul of Geometry Samples
Several Bug Fixes for Geometry Samples


debug.log
```diff
@@ -7,3 +7,11 @@
 [1020/083424.241:ERROR:file_io.cc(135)] WriteFile: The pipe is being closed. (0xE8)
```
Josh-Schifter, line 7: I think you added this file by accident.  Please remove.

src/Components/SampleEditor/Modules.ts
```diff
@@ -0,0 +1,207 @@
+/*---------------------------------------------------------------------------------------------
```
Josh-Schifter, line 1: Please delete this file.  It shouldn't be checked in.  It should be generated at build time and then it will be ignored by .gitignore.

src/Components/SampleGallery/SampleGallery.tsx
```diff
@@ -108,7 +108,7 @@ export class SampleGallery extends React.Component<SampleGalleryProps, SampleGal
   public render() {
     return (
       <>
-        <ExpandableList className="gallery-card-radio" singleExpandOnly={true} singleIsCollapsible={true} defaultActiveBlock={0}>
+        <ExpandableList className="gallery-card-radio" singleExpandOnly={true} defaultActiveBlock={0}>
```
Josh-Schifter, line 111: I think you made this change by accident, possibly through a bad merge.  Please revert.

Josh-Schifter (changes requested):
I'd like for @pmconne to have a look, especially at the changes in GeometryDecorator.tsx.

src/common/GeometryCommon/GeometryDecorator.tsx
```diff
@@ -162,10 +156,59 @@ export class GeometryDecorator implements Decorator {
...
+        }
       } else if (geometry instanceof Path) {
         builder.addPath(geometry);
-      } else if (geometry instanceof Polyface) {
+      } else if (geometry instanceof IndexedPolyface) {
         builder.addPolyface(geometry, false);
+        if (styledGeometry.edges) {
+          // Since decorators don't natively support visual edges,
+          // We draw them manually as lines along each facet edge
+          // Need to offset polyface edges to ensure they don't overlap with the shape itself
```
pmconne, line 184: I think if you supplied the `pickableId` argument to `createGraphicBuilder`, you would not need to mess around with offsetting your edges.

src/common/GeometryCommon/GeometryDecorator.tsx
```diff
@@ -139,9 +139,8 @@ export class GeometryDecorator implements Decorator {
 
   // Iterate through the geometry and point lists, extracting each geometry and point, along with their styles
   // Adding them to the graphic builder which then creates new graphics
-  // TODO: Add the ability to support text rendering
   public createGraphics(context: DecorateContext): RenderGraphic | undefined {
-    const builder = context.createGraphicBuilder(GraphicType.Scene);
+    const builder = context.createGraphicBuilder(GraphicType.Scene, undefined, "GeometryGraphic");
```
pmconne, line 143: This needs to be an Id64String. See documentation for `createGraphicBuilder`.

Josh-Schifter: Would it be correct to use TransientIdSequence.next here?

pmconne: Yes. If `createGraphics` produces new graphics on every call from `decorate`, prefer to obtain and reuse one transient Id; if it caches its graphics for reuse, just obtain a transient Id here.

pmconne: Do edges draw correctly now? Can you add a comment to explain to users why you associated an Id with the graphics?

tomdicarlo: Yes the edges are drawing correctly without the offset after those changes. And yes I can add a comment too.

tomdicarlo: Does the comment I added provide enough information to the user? I don't fully understand why the Id resolves that edge issue, so I'm not sure how much more detail would be necessary in that comment

pmconne: Sorry. I'd say something like "Specifying an Id for the graphics tells the display system that all of the geometry belongs to the same entity, so that it knows to make sure the edges draw on top of the surfaces."

pmconne (approved):

Josh-Schifter (approved):

PR #129, opened 2020-12-08
Thematic update

Ellord207:
These are updates to the existing thematic display sample.  The Sensor functionality will be included in another coming sample.


Ellord207:
@williamkbentley Is the Coffs Harbor Model available in prod?

pmconne:
@Ellord207 I see thematic display applied to terrain in Coffs Harbor but not in Metrostation - why?

pmconne:
@Ellord207 Change mode from height to slope. Switch to different iModel. It displays using height mode.

pmconne:
@Ellord207 you appear to be using the same range for height and slope. This causes e.g. house sample to render almost entirely grey except for some couch cushions. The range for slope is angular - I assume your height range is computed from the project extents Z (which incidentally doesn't work very well for Metrostation).

pmconne:
@Ellord207 Isolines and Stepped With Delimiter are only valid for Height mode.

markschlosseratbentley:
> @markschlosseratbentley can tell you the appropriate range limits for hill shade (that appears not to be documented in ThematicDisplay.ts).

HillShade does not use a range.

pmconne (changes requested):
See comments.

Ellord207:
> @Ellord207 Change mode from height to slope. Switch to different iModel. It displays using height mode though combo box still displays "Slope".

After investigating, the issue is present over the whole showcase.  This will resolved separately from this PR.

Ellord207:
Investigating the issue with "metrostation," I was setting the background settings via a unnecessary cast to ViewState3d and using the 3d display styles.  In metrostation, my changes were somehow being overwritten before rendering.  After changing to only use vp.backgroundMapSettings, the settings I was applying were staying.  Unfortunately, since the fix, I've had frequent crashing while testing that model.  Will continue to investigate.

Ellord207:
> Unfortunately, since the fix, I've had frequent crashing while testing that model. Will continue to investigate
Shortly after experiencing crashing with the Metrostation model, my computer blue screened.  After recovering, I haven't experienced the crashing again.  I'm dropping the investigation into the crashing.

Ellord207:
Due to most models having uninteresting terrain data, or not being good showcases of the thematic display feature, I have removed most all from the list other than the retail build and coffs harbor.

Ellord207:
@pmconne 
> @Ellord207 Isolines and Stepped With Delimiter are only valid for Height mode.

Removed from the select options unless in height mode.

> Slope should limit range to [0..90] degrees.

Range Swaps from project extents to degrees with the modes.

> @Ellord207 Change mode from height to slope. Switch to different iModel. It displays using height mode though combo box still displays "Slope".

This is was an issue with the interaction with the Select React Components from Core.  @tomdicarlo merged a fix for this for the whole Showcase.
@markschlosseratbentley 
>  HillShade does not use a range.

Added a control for sun direction during Hill Shade mode.  I referenced the following website with the correct ranges for the sun direction.  Can you confirm this?
https://pro.arcgis.com/en/pro-app/tool-reference/3d-analyst/how-hillshade-works.htm

pmconne:
> Added a control for sun direction during Hill Shade mode. I referenced the following website with the correct ranges for the sun direction. Can you confirm this?
> https://pro.arcgis.com/en/pro-app/tool-reference/3d-analyst/how-hillshade-works.htm
Talk to @markschlosseratbentley.

markschlosseratbentley:
> > Added a control for sun direction during Hill Shade mode. I referenced the following website with the correct ranges for the sun direction. Can you confirm this?
> > https://pro.arcgis.com/en/pro-app/tool-reference/3d-analyst/how-hillshade-works.htm
> > Talk to @markschlosseratbentley.

@Ellord207 @pmconne Yes this range of values looks good for hillshade.

pmconne (approved):

markschlosseratbentley:
When enabling slope or hillshade, should the background terrain still be toggled on? Slope and hillshade is not currently supported for background terrain and it is therefore not applied in the sample.  It could be a bit surprising to see slope and hillshade not rendering on the terrain at all.  Could we turn off background terrain if those modes are enabled and disallow enabling it?

markschlosseratbentley:
When I enable HillShade, the color scheme is no longer changeable.  It is forever set to whatever value it previously was set to.
Should we instead lock the color scheme to monochrome, which is typically used for HillShade?

image

markschlosseratbentley (changes requested):
See my two comments.

pmconne:
> When enabling slope or hillshade mode, should the background terrain still be toggled on? Slope and hillshade modes are not currently supported for background terrain;

Perhaps the sample could simply note that this is currently not supported. @MarcNeely made some progress toward getting it working

markschlosseratbentley:
> > When enabling slope or hillshade mode, should the background terrain still be toggled on? Slope and hillshade modes are not currently supported for background terrain;
> 
> Perhaps the sample could simply note that this is currently not supported. @MarcNeely made some progress toward getting it working

@pmconne @Ellord207

That sounds like a good resolution to me.

Ellord207:
@markschlosseratbentley 

> Perhaps the sample could simply note that this is currently not supported.

Added note to README about terrain support.

> When I enable HillShade, the color scheme is no longer changeable. It is forever set to whatever value it previously was set to.

I enabled the control for color scheme.  For some reason, I was thinking it was locked to monochrome in iModel.js.

markschlosseratbentley (approved):

BaoTon (approved):

PR #137, opened 2020-12-18
If not specified use the first iModel from spec.

Josh-Schifter:
Fix bug reported by Nathan.


BaoTon (approved):

Ellord207 (changes requested):
Pulled, Started, and navigated to 
"http://localhost:3000/?group=Viewer+Features&sample=classifier-sample&imodel=Retail+Building+Sample"
An alert message appeared and 2 errors were logged to console.

Other case with an unspecified iModel was handled correctly.

Ellord207 (approved):
Looks good.

PR #138, opened 2021-01-25
View Clip Sample - Auto add clip on model switch. Remove stadium.

williamkbentley:
- Auto add clip on model change
- Hide background map
- Remove stadium, too large to be a good example


BaoTon (approved):
Lgtm

PR #141, opened 2021-01-31
Explode sample

Ellord207:
A new sample showing how to create an "Exploded View" effect.


Josh-Schifter:
Just from a user perception point of view the performance of this is not acceptable when I tested it just now.  The animation effect is not smooth at all.  We cannot publish this as it is.

pmconne:
You need to cancel previous async work when queuing new work. e.g., I can slide the slider all the way to 2 and back to 0, but when the animation completes the elements have not returned to their original positions, apparently because you requested a whole bunch of batches of graphics and the order in which they come back is not predictable.

You should keep the same graphics for the same set of elements and simply apply a different transform when the explosion factor changes, instead of requesting brand new graphics every single time the slider value changes slightly. This ought to help with Josh's performance complaints (it will perform much worse in production when it has to talk to the backend over the network).

src/frontend-samples/explode-sample/ExplodeApp.tsx
```diff
@@ -0,0 +1,337 @@
...
+    this.data.forEach((element) => {
+      // TODO: hide elements when Emphasized (not isolated)
+      overrides.overrideElement(element.id, app, true);
+    });
+  }
+
+  /** Apply the supplied function to each [[TileTreeReference]] to be drawn in the specified [[Viewport]]. */
+  public forEachTileTreeRef(viewport: ScreenViewport, func: (ref: TileTreeReference) => void): void {
+    // TODO: update tile tree with new tiles
+    viewport.view.forEachTileTreeRef(func);
```
pmconne, line 281: You are supposed to invoke `func` on the tile tree references owned by **this** provider, not those belonging to the view.

src/frontend-samples/explode-sample/ExplodeApp.tsx
```diff
@@ -0,0 +1,337 @@
...
+  }
+}
+
+/** This provider both hides the original graphics of the element and inserts the transformed graphics. */
+class ExplodeProvider implements TiledGraphicsProvider, FeatureOverrideProvider {
+  private static _instances = new Map<number, ExplodeProvider>();
+  public data: ElementData[] = [];
+  /** Returns a provider associated with a given viewport. If one does not exist, it will be created. */
+  public static getOrCreate(vp: Viewport) {
+    let provider = ExplodeProvider._instances.get(vp.viewportId);
```
pmconne, line 240: Get rid of your map. Use `viewport.findFeatureOverrideProviderOfType<ExplodeProvider>()` instead.

src/frontend-samples/explode-sample/ExplodeApp.tsx
```diff
@@ -0,0 +1,337 @@
...
+import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
+import "common/samples-common.scss";
+import SampleApp from "common/SampleApp";
+import ExplodeUI from "./ExplodeUI";
+import { CoordSystem, DecorateContext, Decorator, EmphasizeElements, FeatureOverrideProvider, FeatureSymbology, GraphicBranch, GraphicType, ImdlReader, IModelApp, SceneContext, ScreenViewport, TileContent, TiledGraphicsProvider, TileRequest, TileTreeReference, Viewport } from "@bentley/imodeljs-frontend";
+import { Matrix3d, Point3d, Point4d, Range3d, Transform, Vector3d } from "@bentley/geometry-core";
+import { ElementGraphicsRequestProps, FeatureAppearance, IModelTileRpcInterface, Placement3d, TileVersionInfo } from "@bentley/imodeljs-common";
+import { ByteStream } from "@bentley/bentleyjs-core";
+
+interface ElementData {
```
pmconne, line 15: This should be split into two interfaces: ElementDataProps to represent the data you get back from your query (and replace Point3d with XYZProps - queries do not return instances of classes); and ElementData to hold the data you create from the query results.

src/frontend-samples/explode-sample/ExplodeApp.tsx
```diff
@@ -0,0 +1,337 @@
...
+    // Limit closest point on sphere to the near plane.
+    viewPt.z = nearViewZ;
+  }
+
+  const viewPt2 = new Point3d(viewPt.x + 1.0, viewPt.y, viewPt.z);
+  return worldToViewMap.transform1.multiplyPoint3dQuietNormalize(viewPt).distance(worldToViewMap.transform1.multiplyPoint3dQuietNormalize(viewPt2));
+}
+
+/** Creates a Range3d containing all the points using the Range3d API. */
+function getRangeUnion(elements: ElementData[]): Range3d {
```
pmconne, line 82: Can't you just union each `ElementData.boundingBox`?
```
const range = Range3d.createNull();
for const elem of elements)
  range.extendRange(elem.boundingBox);
return range;
```

src/frontend-samples/explode-sample/ExplodeUI.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+interface SampleProps {
+  iModelName: string;
+  iModelSelector: React.ReactNode;
+}
+
+interface ExplodeState {
+  isInit: boolean;
+  object: ExplodeObject;
+  explosionFactor: number;
```
pmconne, line 22: What are the units of this - meters?

src/frontend-samples/explode-sample/ExplodeUI.tsx
```diff
@@ -0,0 +1,184 @@
...
+function mapOptions(o: {}): {} {
+  const keys = Object.keys(o).filter((key: any) => isNaN(key));
+  return Object.assign({}, keys);
+}
+
+export default class ExplodeUI extends React.Component<SampleProps, ExplodeState> {
+  public state: ExplodeState;
+  private _objects: ExplodeObject[] = [
+    {
+      name: "Door",
```
pmconne, line 47: What is the point of exploding one element? It has no visible effect.

pmconne (changes requested):
Make your TiledGraphicsProvider actually produce Tiles. Then imodeljs-frontend will take care of scheduling requests for your graphics etc which should fix some of the issues you currently have.

src/frontend-samples/explode-sample/ExplodeUI.tsx
pmconne: Would be interesting to explode the exterior walls of the house to reveal the interior.

Ellord207: > What is the point of exploding one element? It has no visible effect.

The door is only there for testing purposes.  Will remove before completion.  Added TODO: as to not forget.

> Would be interesting to explode the exterior walls of the house to reveal the interior.

Added the room with glass in the back of the house as an object.  Will consider others before completion.

src/frontend-samples/explode-sample/ExplodeApp.tsx
pmconne: please do not resolve other people's comments on your PRs.

pmconne (approved):
"Emphasize" doesn't actually emphasize (outline) the elements. IMO the effect would be more compelling if it opened on the "House Externals" (call it "Exterior" instead?) un-exploded with Emphasis=None and user could click the Explode button. This conveys the idea more clearly than opening on a view looking up at an awkward angle at the half-exploded set of table and chairs floating in space.

You basically can't use Emphasize or None for the interior objects because you can't see them. I would get rid of the table and lamp and replace them with "Windows"; get rid of the "Emphasis" options and instead just always draw the entire model (minus the static graphics for the exploded elements); and have the sample open with the objects unexploded.

src/frontend-samples/explode-sample/ExplodeUI.tsx
```diff
@@ -0,0 +1,224 @@
...
+        <label>Animate</label>
+        <Button onClick={this.onAnimateButton} disabled={this.state.isInit}>{animationText}</Button>
+        <label>Explode Scaling</label>
+        <Slider min={min} max={max} values={[this.state.explodeFactor]} step={step} showMinMax={true} onUpdate={this.onSliderChange} disabled={this.state.isAnimated} />
+        <label>Object</label>
+        <span style={{ display: "flex", justifyContent: "space-between", alignItems: "center" }}>
+          <Select value={this.state.object.name} options={objectEntries} onChange={this.onObjectChanged} style={{ width: "fit-content" }} disabled={this.state.isAnimated} />
+          <Button onClick={this.onZoomButton} disabled={this.state.isInit || this.state.isAnimated}>Zoom To</Button>
+        </span>
+        <label>Emphases</label>
```
pmconne, line 135: /s/Emphases/Emphasis

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+    this.explodeFactor = params.explodeFactor;
+  }
+}
+
+/** Creates a transform matrix translating the range away from the point scaled. */
+function calculateExplodeTransform(centerOfElement: Point3d, displaceOrigin: Point3d, explodeScaling: number) {
+  const vector = Vector3d.createFrom(displaceOrigin);
+  vector.subtractInPlace(centerOfElement);
+  vector.scaleInPlace(explodeScaling - 1);
+  return Transform.createTranslation(vector).inverse()!;
```
pmconne, line 72: Add a comment here explaining that `inverse` will not returned `undefined` because the transform consists solely of translation therefore always invertible.

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+  /** Produce the TileTree corresponding to the specified tree Id. The returned TileTree will be associated with its Id in a Map. */
+  public async createTileTree(id: ExplodeTreeId, iModel: IModelConnection): Promise<ExplodeTileTree | undefined> {
+    if (id.ids.length <= 0)
+      return undefined;
+
+    let info: Promise<TileVersionInfo>;
+    if (this._info === undefined)
+      info = IModelTileRpcInterface.getClient().queryVersionInfo();
+    else
+      info = Promise.resolve(this._info);
```
pmconne, line 106: This is awkward. Why not simply:
```
if (!this._info)
  this._info = await IModelTileRpcInterface.getClient().queryVersionInfo();
```
or query the version info and pass it into the supplier's constructor?

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+    if (explodeScaling === this._prevExplodeFactor) return;
+    this._explodeTransform = calculateExplodeTransform(this.data.boundingBox.center, this.centerOfMass, explodeScaling);
+    this._explodeTransform.multiplyRange(this.data.boundingBox, this._contentRange);
+  }
+
+  /** Output this tile's graphics. */
+  public drawGraphics(args: ExplodeTileDrawArgs) {
+    const gfx = this.produceGraphics();
+    if (!gfx) return;
+    this.setExplodeTransform(args.explodeFactor);
```
pmconne, line 611: This isn't safe - you are not guaranteed that `args` is an instance of your subclass. Can you store the explodeFactor on the TileTree instead?

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+    super({
+      id: `Explode_${params.objectName}_${params.data.length}`,
+      modelId: params.iModel.iModelId!,
+      iModel: params.iModel,
+      priority: TileLoadPriority.Primary,
+      location: Transform.createIdentity(),
+    });
+    this._rootTile = new RootTile(this, params.data, params.tileVersionInfo);
+
+    // Ensures that graphics are not cut off by going outside the iModel extends.
```
pmconne, line 240: /s/iModel extends/project extents

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+
+  public prune(): void {
+    const olderThan = BeTimePoint.now().minus(this.expirationTime);
+    this._rootTile.prune(olderThan);
+  }
+
+  public forcePrune(): void { }
+}
+
+class RootTile extends Tile implements FeatureAppearanceProvider {
```
pmconne, line 300: Explain why this exists.

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+      this.range.extendRange(tile.range);
+    }
+
+    this.loadChildren(); // initially empty.
+    assert(undefined !== this.children);
+
+    this.setIsReady();
+  }
+
+  /** Returns returns it's children elements that contain the graphics for drawing. */
```
pmconne, line 339: "Returns graphics representing the exploded elements."

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+    }
+    return selected;
+  }
+
+  /** Required by FeatureAppearanceProvider. Returns a FeatureAppearance based on element id and other factors or undefined if the it's not to be rendered. */
+  public getFeatureAppearance(source: FeatureAppearanceSource, elemLo: number, elemHi: number, subcatLo: number, subcatHi: number, geomClass: GeometryClass, modelLo: number, modelHi: number, type: BatchType, animationNodeId: number): FeatureAppearance | undefined {
+    const alwaysDrawIds = new Id64.Uint32Set(this._elements.map((element) => element.data.elementId));
+    // If the ids is one of the elements in the tree, always draw it.
+    if (alwaysDrawIds.has(elemLo, elemHi))
+      return FeatureAppearance.fromJSON({});
```
pmconne, line 355: This would be the place to apply the "emphasized" appearance if you keep that option and it's been selected in UI.

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+    return source.getAppearance(elemLo, elemHi, subcatLo, subcatHi, geomClass, modelLo, modelHi, type, animationNodeId);
+  }
+
+  public prune(olderThan: BeTimePoint) {
+    // Never discard ElementTiles - do discard not-recently-used graphics.
+    for (const child of this._elements)
+      child.pruneChildren(olderThan);
+  }
+
+  /** Load this tile's children, possibly asynchronously. Pass them to `resolve`, or an error to `reject`. */
```
pmconne, line 367: Change comment to explain what it *actually* does (nothing) and why.

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+  protected _loadChildren(resolve: (children: Tile[] | undefined) => void, _reject: (error: Error) => void): void {
+    resolve(this._elements);
+  }
+
+  public async requestContent(_isCanceled: () => boolean): Promise<TileRequest.Response> {
+    assert(false, "Root dynamic tile has no content");
```
pmconne, line 373: Replace copy-pasted references to "dynamic tiles".

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+
+  public computeLoadPriority(_viewports: Iterable<Viewport>): number {
+    // We want the element's graphics to be updated as soon as possible
+    return 0;
+  }
+
+  /** Creates an unique id for requesting tiles from the backend. */
+  private makeRequestId(): string {
+    const requestId = requestIdSequence.next();
+    if (requestId.done)
```
pmconne, line 548: `assert` it's not `done` instead of returning non-unique value.

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+      return (-1).toString(16);
+    return requestId.value.toString(16);
```
pmconne, line 550: Add a prefix or suffix to requestId to make it unique among all requests (including, e.g., those made by DynamicIModelTile) and explain why that is necessary.

src/frontend-samples/explode-sample/ExplodeTile.ts
```diff
@@ -0,0 +1,629 @@
...
+  public drawGraphics(args: ExplodeTileDrawArgs) {
+    const gfx = this.produceGraphics();
+    if (!gfx) return;
+    this.setExplodeTransform(args.explodeFactor);
+
+    // Create a new graphics branch we can effect which making change to the rest of the model.
+    const branch = new GraphicBranch();
+    branch.add(gfx);
+
+    // Creates a new set of graphics transformed by the matrix create by the updated explode scaling.  The appearanceProvider insures the elements are drawn;
```
pmconne, line 617: /s/insures/ensures

src/frontend-samples/explode-sample/ExplodeUI.tsx
```diff
@@ -0,0 +1,224 @@
...
+    const vp = this.state.viewport;
+    if (!vp) return;
+    if (this.state.isInit) {
+      this.setState({ isInit: false, emphasize: EmphasizeType.Isolate });
+    }
+    ExplodeApp.refSetData(vp, this.state.object.name, this.state.object.elementIds, this.state.explodeFactor);
+  }
+
+  /** Creates and starts an animator in the viewport. */
+  public createAnimator(): Animator {
```
pmconne, line 85: Might be worth adding this to the list of "purposes of this sample" in README.

src/frontend-samples/explode-sample/ExplodeTile.ts
Ellord207: When originally writing this code, they were generated here so the correct range could be calculated.  This is no longer necessary and the ElementTile generation logic has been moved to the root tile's _loadChildren method.

src/frontend-samples/explode-sample/ExplodeTile.ts
Ellord207: It was organized this way to take advantage of async functions.  As the effects are unnoticeable, I have simplified as suggested.

src/frontend-samples/explode-sample/ExplodeUI.tsx
Ellord207: Referenced in "purpose" of README.

src/frontend-samples/explode-sample/ExplodeTile.ts
Ellord207: Updated.  Please fact check as I could not find documentation on the consequences of matching request ids.

Ellord207:
The Model was updated and the hardcoded ElementIds need to be replaced.  They will eventually be replaced by a query to the iModel.

Ellord207:
> The Model was updated and the hardcoded ElementIds need to be replaced. They will eventually be replaced by a query to the iModel.

Hard coded element ids have been replaced or removed.  Will revisit objects when model is updated.

src/frontend-samples/explode-sample/ExplodeApp.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+  /** Enables an animator using the Viewport API. If the animator is undefined, any active animator will be removed. */
+  public static setAnimator(vp: Viewport, animator?: Animator) {
+    vp.setAnimator(animator);
+  }
+}
+
+/** This provider both hides the original graphics of the element and inserts the transformed graphics. */
+class ExplodeProvider implements TiledGraphicsProvider, FeatureOverrideProvider {
+  // These methods support the provider as an object.
```
Josh-Schifter, line 119: // These methods support the provider as an object.

That comment seems entirely pointless.  Right?

src/frontend-samples/explode-sample/ExplodeApp.tsx
```diff
@@ -0,0 +1,178 @@
...
+    this.vp.dropFeatureOverrideProvider(this);
+    this.vp.dropTiledGraphicsProvider(this);
+  }
+
+  /** Signals to the viewport that this provider needs to be recalculated. */
+  public invalidate() {
+    this.vp.setFeatureOverrideProviderChanged();
+  }
+
+  // These methods support the functionality of the provider.
```
Josh-Schifter, line 157: // These methods support the functionality of the provider.

I suggest removing this comment

Josh-Schifter (commented):
Can we do something about the default view for the sunroom?  If we spin the view around to view the house from the back, then I think the sunroom might make a better default.

Likewise, 'zoomto' for the sunroom doesn't work well since it's blocked by the house.

Ellord207:
> Can we do something about the default view for the sunroom? If we spin the view around to view the house from the back, then I think the sunroom might make a better default.
> 
> Likewise, 'zoomto' for the sunroom doesn't work well since it's blocked by the house.



> Can we do something about the default view for the sunroom? If we spin the view around to view the house from the back, then I think the sunroom might make a better default.
> 
> Likewise, 'zoomto' for the sunroom doesn't work well since it's blocked by the house.

Added a standard rotation when zooming, isolated the sunroom, and set as default.

Josh-Schifter (approved):

PR #148, opened 2021-02-12
Screen-space effects sample

pmconne:
3 effects: lens distortion, saturation, and vignette. All 3 can be customized via sliders.


src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+      // Ask the viewport to redraw its contents so that the changes to the effect are immediately visible.
+      if (this.state.viewport)
+        this.state.viewport.requestRedraw();
+    };
+
+    return (
+      <>
+        <span>{label}</span>
+        <input type="range" min={min} max={max} step={step} value={value} onChange={updateValue}></input>
```
Ellord207, line 54: onChange only raises an event when the slider is finished moving.  Generally, onUpdate is better since it raises an event at every step, since the interaction is smoother.

Ellord207 (changes requested):

src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
pmconne: I'm not using a `Slider`, I'm using an `input`. I switched from `onChange` to `onInput`, which is supposed to update whenever the value changes, but it still seems to only invoke callback when slider stops moving.

pmconne:
@Ellord207 thumbnails acceptable?

src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
Ellord207: Just noticed you're using a basic html element and not ui-core Slider component.  Pushing fix.

Ellord207:
> @Ellord207 thumbnails acceptable?

They are both the correctly formatted and sized.

But I do have a complaint.  Typically, the icon matches the view of the sample when it is first loaded.

pmconne:
> > @Ellord207 thumbnails acceptable?
> 
> They are both the correctly formatted and sized.
> 
> But I do have a complaint. Typically, the icon matches the view of the sample when it is first loaded.

That explains why most of the thumbnails look identical. I wanted to actually show the effect.

src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
Ellord207: > > > @Ellord207 thumbnails acceptable?
> > 
> > 
> > They are both the correctly formatted and sized.
> > But I do have a complaint. Typically, the icon matches the view of the sample when it is first loaded.
> 
> That explains why most of the thumbnails look identical. I wanted to actually show the effect.

You can limit the list of selectable iModels, then when loading the view, select a customized camera position as you like it.

Ellord207:
> > > @Ellord207 thumbnails acceptable?
> > 
> > 
> > They are both the correctly formatted and sized.
> > But I do have a complaint. Typically, the icon matches the view of the sample when it is first loaded.
> 
> That explains why most of the thumbnails look identical. I wanted to actually show the effect.

You can limit the list of selectable iModels, then when loading the view, select a customized camera position as you like it.

src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
Ellord207: Pushed fix, but it's still not as smooth as I would expect.  Is this just it's performance or would there be a way to demand a redraw quicker?

pmconne:
> > > > @Ellord207 thumbnails acceptable?
> > > 
> > > 
> > > They are both the correctly formatted and sized.
> > > But I do have a complaint. Typically, the icon matches the view of the sample when it is first loaded.
> > 
> > 
> > That explains why most of the thumbnails look identical. I wanted to actually show the effect.
> 
> You can limit the list of selectable iModels, then when loading the view, select a customized camera position as you like it.

I don't like any of these iModels. Is there a project to which I can upload a new one?

Ellord207:
> > > > > @Ellord207 thumbnails acceptable?
> > > > 
> > > > 
> > > > They are both the correctly formatted and sized.
> > > > But I do have a complaint. Typically, the icon matches the view of the sample when it is first loaded.
> > > 
> > > 
> > > That explains why most of the thumbnails look identical. I wanted to actually show the effect.
> > 
> > 
> > You can limit the list of selectable iModels, then when loading the view, select a customized camera position as you like it.
> 
> I don't like any of these iModels. Is there a project to which I can upload a new one?

@Josh-Schifter, @williamkbentley (Will handles the iModels in the sample) Both are in a hackathon for today and will be free start of next week.

src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
pmconne: `Viewport.requestRedraw` is supposed to request an animation frame. It doesn't. I pushed a workaround until that's fixed and fix integrated into the showcase.

pmconne:
@Ellord207 I updated the thumbnails. If at some point we can get @Josh-Schifter to let us use more aesthetically pleasing model(s) I will update this sample (and the display styles sample) to do so.

src/frontend-samples/screen-space-effects-sample/ScreenSpaceEffectsUI.tsx
```diff
@@ -0,0 +1,153 @@
...
+    // Were the settings controlling the effects changed?
+    if (!equalEffectsConfigs(this.state.effectsConfig, prevState.effectsConfig)) {
+      changed = true;
+      updateEffectsConfig(this.state.effectsConfig);
+    }
+
+    if (changed) {
+      viewport.requestRedraw();
+
+      // ###TODO requestRedraw is supposed to do this, but currently doesn't. Remove once that is fixed.
```
williamkbentley, line 118: In what version will this be included?

pmconne: Fix is currently on master so 2.13. Shall I backport to 2.12?

williamkbentley: That is not necessary. Just wanted to know when we would need to circle back to this TODO.

williamkbentley (approved):

pmconne:
@Ellord207 satisfied with changes?

Ellord207 (approved):

PR #151, opened 2021-02-16
Tweak display styles sample.

pmconne:
- Default to villa model for which these styles were created.
- Default to sun-dappled style. Illustration was first choice but looks bad at the 740x450 thumbnail resolution.
- Replace bright sunny skybox for 'moonlit' style.
- Update thumbnails.


williamkbentley (approved):
LGTM

PR #153, opened 2021-02-24
Common reorganization/sample refactor

tomdicarlo:
This branch makes several changes to all of the samples as well as the showcase in order to prepare existing showcase samples to be ported over to the sandbox.

Primary Changes:
-Reorganized the common and Components folders. common now contains code that is shared between several samples(such as the geometry decorator), while Components contain code that is used directly by the showcase(such as the sample gallery)
-Removed the need for samples to have setup and teardown functions. Code that was used by samples as part of setup or teardown has been moved into ComponentDidMount and ComponentWillUnmount on the UI Components.


debug.log
```diff
@@ -0,0 +1,2 @@
+[1222/133038.972:ERROR:directory_reader_win.cc(43)] FindFirstFile: The system cannot find the path specified. (0x3)
```
williamkbentley, line 1: Delete this file

williamkbentley:
LGTM; Tested the samples I created as well as several others. Also quickly scanned though code and looked good as well.

williamkbentley (approved):

StefanRetief (approved):

PR #158, opened 2021-03-04
updated monaco-editor to latest

StefanRetief:
Upgrade for monaco editor. Fixed various issues:

- Fixed split-screen for "Problems" pane not rendering correctly
- Fixed transitions for opening & closing gallery/editor
- fixed minor styling issues
- Refactored markdown viewer into smaller components

**NOTE:** The sample editor component has been refactored into a react hook component. The editor has the ability to hook into a class component, but the work required is much too convoluted when hooks work fine.

I tested some samples and it appears that the files compile fine. Please do let me know if there are any samples that you test and run into a file not compiling as it should.

PR #160, opened 2021-03-05
Advanced HTML Markers sample

MichaelBelousov:
please let me know if cat videos/pictures are not what we want for showing off HTML markers
- three modes you can toggle in controls: Picture, Video, Form
- can place markers, remove all markers
- image and video markers scale with the view, but this has to be implemented manually

it's also possible that we want to separate an example of `useMarker/MarkerComponent` from `@bentley/imodel-react-hooks` into another sample.

Default view (form):
image

image view:
image

Video view:
image


src/sampleManifest.tsx
```diff
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
-* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
-*--------------------------------------------------------------------------------------------*/
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
MichaelBelousov, line 2: looks like my prettier is being aggressive, I'll revert my other changes to this file

williamkbentley:
A few errors in "Problems" tab:
image

utils.tsx can be added to sampleSpec. Unsure about the other errors, possible I didn't build correctly. @StefanRetief?

src/frontend-samples/html-marker-sample/HtmlMarkerUI.tsx
```diff
@@ -0,0 +1,124 @@
...
+import ToolsProvider, { ToolsContext } from "./HtmlMarkerTools";
+import { SandboxViewport } from "common/SandboxViewport/SandboxViewport";
+import { ControlPane } from "common/ControlPane/ControlPane";
+import { IModelJsViewProvider } from "@bentley/imodel-react-hooks";
+import HtmlMarker from "./HtmlMarker";
+
+const i18nNamespaceKey = "react-marker-i18n-namespace" as string;
+
+export enum HtmlContentMode {
+  Image = "picture",
```
williamkbentley, line 25: Can you capitalize values?

src/frontend-samples/html-marker-sample/HtmlMarkerUI.tsx
```diff
@@ -0,0 +1,124 @@
...
+          )}
+        </ToolsContext.Consumer>
+      </div>
+    </>
+  );
+
+  return (
+    <ToolsProvider i18nNamespace={sampleNamespace} addMarker={addMarker}>
+      <ControlPane
+        instructions="Use the options below to control the marker pins.  Click a marker to open a menu of options."
```
williamkbentley, line 105: I think you forgot to update these instructions.

MichaelBelousov:
> A few errors in "Problems" tab:
did you `npm install` the new dependency?

src/frontend-samples/html-marker-sample/README.md
```diff
@@ -0,0 +1,21 @@
+# HTML Marker Sample
+
+Copyright © Bentley Systems, Incorporated. All rights reserved.
+
+This sample demonstrates how to use the [@bentley/imodel-react-hooks](/https://github.com/imodeljs/viewer-components-react/tree/master/packages/imodel-react-hooks)
```
williamkbentley, line 5: Leading slash on @bentley/imodel-react-hooks link causes broken link

src/frontend-samples/html-marker-sample/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+The purpose of this sample is to demonstrate the following:
+
+- Using the [Marker component or useMarker hook](https://github.com/imodeljs/viewer-components-react/tree/master/packages/imodel-react-hooks#usemarker) to make
+  markers that render arbitrary dynamic JSX with React.
+- Advanced React state integration with a [PrimitiveTool](https://www.imodeljs.org/reference/imodeljs-frontend/tools/primitivetool/) that can use all React state
+
+## Description
+
+[Markers](https://www.imodeljs.org/learning/frontend/markers/) are a common technique to draw a user's attention to a particular spatial location within a digital twin.  Markers are typically organized into [marker sets](https://www.imodeljs.org/reference/imodeljs-frontend/views/markerset/) and displayed by a [decorator](https://www.imodeljs.org/reference/imodeljs-frontend/views/decorator/). However, [@bentley/imodel-react-hooks](https://github.com/imodeljs/viewer-components-react/tree/master/packages/imodel-react-hooks) is a
```
williamkbentley, line 19: Would be nice to link to Marker sample here. After describing about markersets and decorators: "The Marker Pin sample demonstrates this setup"

src/sampleManifest.tsx
```diff
@@ -89,6 +90,7 @@ export const sampleManifest: SampleSpecGroup[] = [{
     getMarkerPinSpec(),
     getMultiViewportSpec(),
     getPropertyFormattingSpec(),
+    getHtmlMarkerSpec(),
```
williamkbentley, line 93: Any reason you put it here? This list is mostly alphabetical; although this one may be better off near MarkerPin

williamkbentley (dismissed):

src/sampleManifest.tsx
MichaelBelousov: good point... originally I had named it "react markers sample" which is why it was after `P`. I'll move it to MarkerPin.

williamkbentley:
> > A few errors in "Problems" tab:
> 
> did you `npm install` the new dependency?

Yeah, I have it installed. The editor imports the typedefs, but I'm not quite sure where. Stefan will probably be able to help.

MichaelBelousov:
> For the thumbnail, we typically like the thumbnail to reflect the default state of the sample. In this case it should be a form (Or change the default HTML maker to image)

I think I will switch to the default being the video tag and get a thumbnail of that.

StefanRetief:
> A few errors in "Problems" tab:

@bentley/imodel-react-hooks needs to be added to the "modules.json" file, like so:

```json
{
 "name": "@bentley/imodel-react-hooks",
 "import": "@bentley/imodel-react-hooks"
}
```

The error in the problems pane for this issue won't be fixed until PR #158 is completed:

The ./utils dependencies need to be added to the SampleSpec if you would like them to be editable and compilable:

```
{
 name: "utils.ts",
 import: import("!!raw-loader!./utils.ts"),
},
```

Or added to the "modules.json" and imported as a non-relative file relative to the src folder if just making them viewable is preferred. Typically, its best if "utils.ts" is moved to common if this is desired:

_In "HtmlMarker.tsx":_  ```import { clamp, HtmlMarkerWrapper } from "common/utils";```

_In "modules.json":_
```json
{
 "name": "common/UIComponents/UIComponentContainer",
 "import": "./src/common/UIComponents/UIComponentContainer",
 "typedef": true
},
```

MichaelBelousov:
@Josh-Schifter could you make recommendations on changes to the HTML content or "narrative"? I now remember you thought directly showing videos might not be a good idea (the original idea of using a video tag came from a colleague, on being able to see live CCTV monitoring footage of infrastructure). I think a better "narrative" would make this sample much more useful.

Josh-Schifter:
Some ideas:
- My original thinking is that it should work like the IoT Alert sample where the user first selects an element, and then we put the marker on the center of the element.
- I don't like that changing the picker affects all the existing markers.  You should be able to create different types of markers and have them exist at the same time.
- The most compelling type of HTML marker is probably just simple text which embeds real-time data like we did in the Doosan IoT proof-of-concept.  I also prefer the styling we used there which was light text on a dark grey background.  This is all to contrast with your 'Form' marker.
- Images showing some industrial equipment would probably be better than cats.
- Interactive content like video and forms does seem too 'heavy' for markers that might be 'in the background'.
    - For the Doosan case clicking on the marker brings that turbine 'forward' so it becomes the focus.  When in focus then we show additional markers... maybe that's the way to get to the interactive ones.

MichaelBelousov:
My thinking is I'll use the Doosan IoT PoC style, and change the tool to gather some properties of the selected element, creating a dynamic marker on element press which is removed when clicking off the element. Maybe some small icon to indicate we can render more than just text. At that point I can't think of any useful controls besides:
- description saying: "click on an element to see a marker describing it"
- remove all markers control

MichaelBelousov:
I don't have time for this right now, I'm going to close it.
I would however say that used in this PR is a particularly well-suited technique for closely and idiomatically integrating React state with our class-based API for markers. It's described [here](https://medium.com/itwinjs/provider-local-class-pattern-dc44bab33144). 
I think it's worth having a sample with this technique so I'd like to revive this as an "advanced markers in React Apps" sample eventually.

PR #165, opened 2021-03-09
Upgrade to 2.12.3 to fix immer audit issue.

BaoTon:


BaoTon:
I don't think this is the correct approach - no-unused-vars is now a warning instead of an error.
I've basically just disabled eslint from the react-app side, but I don't think this approach is desired.

Setting the environment variable DISABLE_ESLINT=true didn't do the trick... @calebmshafer  any other suggestions?
I'm trying to make a better overrides that doesn't completely ruin the eslint.

BaoTon:
Closing in favor of #166

PR #168, opened 2021-03-10
Loading optimization

linasburneika:
Refactored app for lazy loading. This PR depends on "updated monaco-editor to latest", which needs to be merged first.
The main SampleShowcase component was changed to FunctionComponent style, which caused a major code rewrite.
sampleSpec was changed in every sample as well - it has function callbacks to get editor files and readme file, which are lazyly loaded on sample click in gallery.


StefanRetief:
@linasburneika #158 has been merged, This should be unblocked now

StefanRetief (approved):
LGTM

StefanRetief (changes requested):
@linasburneika Something is causing the entire sample showcase to rerender on state change. For example, when the user starts dragging on a pane or closes a pane, that causes both the editor and visualizer to rerender as well, even though neither are reliant on those state changes.

StefanRetief (approved):

williamkbentley (approved):
I tested every sample, and looks good. I uncovered about a half dozen bugs unrelated to these changes. I'll create Issues for those. But this looks good to me.

BaoTon (approved):

PR #172, opened 2021-03-15
Editor update

StefanRetief:


src/frontend-samples/clashreview-sample/sampleSpec.ts
```diff
@@ -18,6 +18,7 @@ export function getClashReviewSpec(): SampleSpec {
       { name: "ClashDetectionApis.ts", import: import("!!raw-loader!./ClashDetectionApis.ts") },
       { name: "ClashTable.tsx", import: import("!!raw-loader!./ClashTable.tsx") },
       { name: "MarkerPinDecorator.ts", import: import("!!raw-loader!frontend-samples/marker-pin-sample/MarkerPinDecorator.ts") },
+      { name: "PopupMenu.tsx", import: import("!!raw-loader!frontend-samples/marker-pin-sample/PopupMenu.tsx") },
```
Josh-Schifter, line 21: I'm pretty sure that @wilmaier omitted PopupMenu intentionally.

wilmaier: Well yes but I forgot it was needed to resolve the reference in the editor.  I actually talked to Stefan about adding it back.

Josh-Schifter: I'll approve for the short term, but we really should refactor so there is no direct dependence from MarkerPin to PopupMenu.

Josh-Schifter (approved):

PR #178, opened 2021-03-17
Metrostation categories hidden

Ellord207:
This change extends the View Setup features to also hide coincident geometry in the metrostation model.

I've attempted to not have the whole file marked as changed, but without success.


src/api/viewSetup.ts
```diff
@@ -1,130 +1,144 @@
...
+      const selectInCategories = `SELECT ECInstanceId FROM bis.Category WHERE CodeValue IN ('${categoryCodes.join("','")}')`;
+      for await (const row of imodel.query(selectInCategories))
+        ids.push(row.id);
+    };
+    if (imodel.name === "house bim upload")
+      // The callout graphics in the house model are ugly - don't display them.
+      await addIdsByCategory("Callouts");
+
+    if (imodel.name === "Metrostation2")
+      // There is lots of coincident geometry. Removes the more visible instances.
```
williamkbentley, line 110: Grammar: "There is coincident geometry. Remove the more visible instances."

Josh-Schifter (approved):

PR #179, opened 2021-03-18
Support for iTwin Viewer

StefanRetief:
Provides support for iTwin viewer and reloadable viewport. Any sample that takes advantage of itwin viewer will need to add the 'iTwinViewerReady" property to their sample spec. This will notify the visualizer component to *not* setup the imodelApp before running the visualizer.

**NOTE:** The biggest change is a rename in the sample spec from 'customModelList" to "modelList" (about 22 files).

Other changes include the @itwin-sandbox folder and components. This folder serves as a common module between the sandbox and the showcase. Everything that changes here (minus some things to do with URL params) will also be changed in the sandbox. 

**NOTE:** Wait until imodeljs 2.14 to merge.


src/index.tsx
```diff
@@ -7,14 +7,19 @@ import React from "react";
...
 import "./index.scss";
 import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
 
-ReactDOM.render(
-  <App />,
-  document.getElementById("root"),
-);
+(async () => {
+  await FeatureToggleClient.initialize();
+  await AuthorizationClient.initializeOidc();
```
linasburneika, line 17: OIDC init needs to be moved just before imodel starts loading. Token is not needed for gallery and editor display. Obtaining a token takes up to 1 second, which increase white screen delay before app load.

src/index.tsx
```diff
@@ -7,14 +7,19 @@ import React from "react";
...
+import { AuthorizationClient } from "@itwinjs-sandbox";
 import "./index.scss";
 import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
 
-ReactDOM.render(
-  <App />,
-  document.getElementById("root"),
-);
+(async () => {
+  await FeatureToggleClient.initialize();
```
linasburneika, line 16: Should be moved downstream before it's usage. Adds ~150ms to app statup delay. Not sure if feature toggling is really needed at all. Can anyone comment?

src/index.tsx
```diff
@@ -7,14 +7,19 @@ import React from "react";
 import ReactDOM from "react-dom";
 import { App } from "./Components/App/App";
 import * as serviceWorker from "./serviceWorker";
-
+import { FeatureToggleClient } from "FeatureToggleClient";
+import { AuthorizationClient } from "@itwinjs-sandbox";
```
linasburneika, line 11: Because of imports in AuthorizationClient, it drags in complete @bentley/itwin-client and some other deps to initial app bundle, causing it to bloat again to 5.5Mb

src/SampleSpec.ts
```diff
@@ -1,3 +1,4 @@
+import { SampleIModels } from "@itwinjs-sandbox";
```
linasburneika, line 1: Extract this enum into separate file without any imports to be used here. Those imports drags in megabytes of dependencies to initial bundle size. The same applies to sampleSpec files where this interface is used

src/@itwinjs-sandbox/authentication/AuthorizationClient.ts
```diff
@@ -0,0 +1,93 @@
...
+      ...await body,
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      _userInfo: { id: "MockId" },
+      // eslint-disable-next-line @typescript-eslint/naming-convention
+      _tokenString: body._jwt,
+    };
+    this._accessToken = AccessToken.fromJson(tokenJson);
+
+    // Automatically renew if session exceeds 55 minutes.
+    setTimeout(() => {
```
linasburneika, line 79: Probably setInterval is better here, to renew it continously.

BaoTon: Isn't this a recursive function? setInterval will cause a stacking effect if calling this.generateTokenString().

We can change this to setInterval if we call it outside the function.

src/index.tsx
StefanRetief: Move OIDC init to Sample Visualizer component

src/index.tsx
StefanRetief: Move Feature Toggle client init to Tab Nav (it's the only place that it is getting used).

src/index.tsx
StefanRetief: Removed imports

src/SampleSpec.ts
StefanRetief: Moved SampleIModels to its own file and changed import statements to reference the file directly.

src/frontend-samples/viewport-only-sample/ViewportOnlyUI.tsx
```diff
@@ -2,23 +2,66 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
-import * as React from "react";
```
StefanRetief, line 5: @Josh-Schifter Can you comment on this sample code?

Josh-Schifter: It is a lot more code than before.  Was 24 lines, now 67 lines.  That is unfortunate but I don't see an obvious way to condense it.

src/@itwinjs-sandbox/authentication/AuthorizationClient.ts
```diff
@@ -0,0 +1,98 @@
...
+  private static _initialized: boolean = false;
+
+  constructor() {
+    this.onUserStateChanged = new BeEvent();
+  }
+
+  public static async initializeOidc(): Promise<void> {
+    if (!this._initialized) {
+      const authClient = new AuthorizationClient();
+      const userURL = Config.App.get("imjs_sample_showcase_user", "https://prod-imodeldeveloperservices-eus.azurewebsites.net/api/v0/sampleShowcaseUser");
```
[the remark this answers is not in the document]
BaoTon: Do you mean hide the URL behind something like BUDDI? 

The user is globally accessible client-side and requires no permission to access, we could take an additional measure in obfuscating the URL but anyone who wants to know the address will know it regardless of the measure taken.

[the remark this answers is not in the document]
aruniverse: I know this is cause we don't have public imodels and this is effectively so there's no need to sign in.
I dont love that we're exposing an endpoint users can hit to get a token, but i guess it scoped enough so the token is only valid for the test org account and can only access those imodels. 

I cant think of a better solution for this atm. You're right it'll be globally accessible client side so if i wanted to inspect the network tab while having the showcase up, i could deduce what endpoint to hit myself.

i'm guessing you guys already discussed this with the security team and theyre ok with this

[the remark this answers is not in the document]
BaoTon: That's correct, this is hopefully resolved in the future with a new scoping mechanism.

src/frontend-samples/viewer-only-2d-sample/ViewerOnly2dUI.tsx
```diff
@@ -2,125 +2,106 @@
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
StefanRetief, line 2: @Josh-Schifter I've made some changed to this sample (ViewerOnly2d) based on the discussion we had previously. I've modified it to not rely solely on the widget. Could you review it?

src/frontend-samples/viewer-only-2d-sample/ViewerOnly2dUI.tsx
```diff
@@ -7,19 +7,35 @@ import React from "react";
 import { Viewer } from "@bentley/itwin-viewer-react";
 import { ControlsWidget } from "./ViewerOnly2dWidget";
 import { IModelConnection, ViewState } from "@bentley/imodeljs-frontend";
+import ViewerOnly2dApp from "./ViewerOnly2dApp";
+import { ModelProps } from "@bentley/imodeljs-common";
+export interface TwoDState {
+  drawingElements: JSX.Element[];
+  sheetElements: JSX.Element[];
+  sheets: ModelProps[];
+  drawings: ModelProps[];
```
Josh-Schifter, line 16: It seems redundant for the state to have both the elements and the modelprops.  I suggest only keeping the modelprops here and generating the jsx.elements in the widget as needed.  Those are only used for building the Select in the widget.

Also, I don't like the name 'TwoDState'.  This is part of the state so it doesn't need the word 'state'.  How about 'ModelLists'?

src/frontend-samples/viewer-only-2d-sample/ViewerOnly2dUI.tsx
```diff
@@ -34,13 +50,36 @@ export default class ViewportOnly2dUI extends React.Component<{}, ViewportOnly2d
   private _getSampleUi = (iModelName: SampleIModels) => {
     return new SampleWidgetUiProvider(
       "The picker below shows a list of 2D models in this iModel.",
-      <ControlsWidget />,
+      <ControlsWidget twoDState={this.state.twoDState} />,
       { iModelName, onIModelChange: this._changeIModel }
     );
   };
 
+  private _getDrawingModelList = (models: ModelProps[]) => {
```
Josh-Schifter, line 58: I wouldn't have functions for _getDrawingModelList and _getSheetModelList.  Rather this code would be directly in the widget.  My reason is because all this building of options is just a UI detail, and not an important part of the sample.

src/frontend-samples/viewer-only-2d-sample/ViewerOnly2dUI.tsx
```diff
@@ -34,13 +50,36 @@ export default class ViewportOnly2dUI extends React.Component<{}, ViewportOnly2d
...
+    models.forEach((model: ModelProps, index) => {
+      sheetViews.push(<option key={`${index}sheet`} value={`${index}sheet`}>{model.name}</option>);
+    });
+    return sheetViews;
+  };
+
   private _oniModelReady = (iModelConnection: IModelConnection) => {
-    ViewSetup.getDefaultView(iModelConnection)
+    ViewerOnly2dApp.get2DModels(iModelConnection)
+      .then(async (result) => {
```
Josh-Schifter, line 76: Why do you use .then and not await?  Await seems more common.

Josh-Schifter (commented):
I'm still hesitant about the separate file for the Widget class.  It seems to me that draws unnecessary attention to it.  If we moved the Widget into UI.tsx (at the bottom) then it would lump together with the other UI code and I would expect most users wouldn't need to look at it.

src/@itwinjs-sandbox/components/imodel-selector/IModelSelector.tsx
```diff
@@ -0,0 +1,41 @@
...
+import React, { FunctionComponent } from "react";
+import { Select } from "@bentley/ui-core";
+import "@bentley/icons-generic-webfont/dist/bentley-icons-generic-webfont.css";
+import { SampleIModels } from "@itwinjs-sandbox/SampleIModels";
+
+// The Props and State for this sample component
+interface IModelSelectorProps {
+  iModelNames: SampleIModels[];
+  iModelName: SampleIModels;
+  onIModelChange: (iModelName: SampleIModels) => void;
```
aruniverse, line 14: should this be async?

StefanRetief: Not necessarily. At the end of the day, this component is just a select component. It would probably be better served as a regular function for reusability's sake.

package.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "@bentley/bentleyjs-core": "2.13.0",
     "@bentley/context-registry-client": "2.13.0",
+    "@bentley/frontend-application-insights-client": "^2.13.1",
```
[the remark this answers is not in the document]
StefanRetief: Expecting 2.14 to be merged in first. If it's still misaligned, I'll fix it then.

src/SampleSpec.ts
```diff
@@ -22,7 +23,7 @@ export interface SampleSpec {
   image: string;
   readme?: () => Promise<{ default: string }>;
   files?: () => SampleSpecFile[];
-  customModelList?: string[];
-  sampleClass?: typeof React.Component;
+  modelList?: SampleIModels[];
```
Josh-Schifter, line 26: As long as we are changing this, it should be iModelList with the leading "i".

src/frontend-samples/emphasize-elements-sample/EmphasizeElementsWidget.tsx
```diff
@@ -0,0 +1,243 @@
...
+        <span>Override</span>
+        <ColorPickerButton initialColor={colorValueState} onColorPick={_onColorPick} disabled={emphasizeElementsProps.selectionIsEmpty} />
+        <Button buttonType={ButtonType.Primary} onClick={() => emphasizeElementsProps.handleActionButton(ActionType.Override)} disabled={emphasizeElementsProps.selectionIsEmpty}>Apply</Button>
+        <Button buttonType={ButtonType.Primary} onClick={() => emphasizeElementsProps.handleClearButton(ActionType.Override)} disabled={!emphasizeElementsProps.overrideIsActive}>Clear</Button>
+      </div>
+    </>
+  );
+};
+
+/// ///////////////////////////////////////////////
```
Josh-Schifter, line 73: Lots of junk here.

PR #182, opened 2021-03-23
Snow Sample

Ellord207:
A new sample featuring the particle collection builder used in a decorator to create snow and rain effects in the view.


Ellord207:
@pmconne, I want your opinion on the rain particle.  I have included 2 different rain particles with this PR, but I'm not sure which to use.  "particle_rain2.png" looks the best normally, but when the wind is strong, can look silly as it's more vertical.  I replaced it with the circular gradient, "particle_rain.png".

I would like to confirm with you that there is no way to programmatically transform the texture.  My understanding is that the correct way to "rotate" a texture would be to prepare the texture at different angles before compile then load the specific angle you need during run time.

pmconne:
> I would like to confirm with you that there is no way to programmatically transform the texture. My understanding is that the correct way to "rotate" a texture would be to prepare the texture at different angles before compile then load the specific angle you need during run time.

Normally you could rotate the quads, or adjust their UV coords, but ParticleCollectionBuilder doesn't give you control over that. You could use the CanvasRenderingContext2d APIs to produce a rotated version of your texture image. I'd just go with selecting the alternate image if the wind is strong enough to make the default image look weird.

pmconne (changes requested):

src/frontend-samples/snow-decoration-sample/sampleSpec.ts
```diff
@@ -0,0 +1,23 @@
...
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+
+import { SampleIModels } from "common/IModelSelector/IModelSelector";
+import { SampleSpec } from "SampleSpec";
+
+export function getParticleSnowSampleSpec(): SampleSpec {
+  return ({
+    name: "snow-sample",
+    label: "Particle Effect (Snow)",
```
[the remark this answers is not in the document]
Ellord207: "weather" is more broad than the scope of this sample.
I'd also argue "Particle Sample (Precipitation)" while accurate, isn't as catchy as "Particle Sample (Snow)"
@Josh-Schifter, has already okayed the name, and I'll leave the final call up to him.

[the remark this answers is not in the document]
Josh-Schifter: I suggested "Particle Effect (Snow & Rain)" or some variant if the '&' is a problem.  I want the name to be search friendly and I don't foresee anyone searching for 'precipitation'.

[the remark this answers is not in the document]
Ellord207: No issue with the "&" in the label, but I had issues changing the path name.  I went with "snow-rain-sample".

Ellord207:
> I'd just go with selecting the alternate image if the wind is strong enough to make the default image look weird.

I did go with this method.  It looks much better than the static gradient which I left as another option.

src/frontend-samples/snow-decoration-sample/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+For a basic decorator, the [Decorator](https://www.itwinjs.org/reference/imodeljs-frontend/views/decorator/) interface is implemented. The decorate method adds the [RenderGraphics](https://www.itwinjs.org/reference/imodeljs-frontend/rendering/rendergraphic/) to the given [DecorateContext](https://www.itwinjs.org/reference/imodeljs-frontend/rendering/decoratecontext). Once the decorator is created, pass it to the [ViewManager.addDecorator](https://www.itwinjs.org/reference/imodeljs-frontend/views/viewmanager/adddecorator/) method to have it rendered in all active views. This sample is specifically for working with particle effects. For examples of other decorators, see [Heatmap Decorator Sample](https://www.itwinjs.org/sample-showcase/?group=Viewer+Features&sample=heatmap-decorator-sample&imodel=Metrostation+Sample) and this sample's [HighlightEmitter](./FireDecorator.ts).  For another sample using particles in a decorator, see [Particle Decoration (Fire)](https://www.itwinjs.org/sample-showcase/?group=Viewer+Features&sample=fire-sample).
```
Josh-Schifter, line 18: The reference to Highlight Emitter should be removed.

In my opinion, this works better for linking to a different sample: [volume](../volume-query-sample/readme.md).  I like it since it doesn't reload the whole page.

src/frontend-samples/snow-decoration-sample/README.md
```diff
@@ -0,0 +1,28 @@
...
+## Purpose
+
+The purpose of this sample is to demonstrate the following:
+
+* Creating a decorator and rendering it in the active view.
+* Working with particle effects using the [ParticleCollectionBuilder](https://www.itwinjs.org/reference/imodeljs-frontend/rendering/particlecollectionbuilder) API.
+
+## Description
+
+This sample provides an example of a particle effect in view space (as opposed to world space).  The sample provides features to configure a decorator emitting snow particles.  The decorator is creating using predefined props defined in [SnowDecorationApp.ts](./SnowDecorationApp.tsx).  Select which effect is used with the drop down menu.
```
Josh-Schifter, line 16: ...emitting snow and rain...

The decorator is *created* using redefined props

src/frontend-samples/snow-decoration-sample/README.md
```diff
@@ -0,0 +1,28 @@
...
+The texture for each particle is loaded by the sample, and owned by the decorator.  When the sample disposes, it signals the decorator is also dispose of the texture.
+
+For this sample, each particle is updated before being rendered.  All updates are scaled by a time delta to smooth any inconsistent times between updates.  To ensure the decorators are always rendered by the viewport, an event listener is added to the [Viewport.onRender](https://www.itwinjs.org/reference/imodeljs-frontend/views/viewport/?term=onrender#onrender) event that invalidates the decorators.  This will force them to re-render with each frame.
+
+Notes:
+
+* This is in no way a simulation.  The weather presented only serves as a visual effect.
+* The APIs related to particles are in beta.  Please consulate the latest documentation before adapting for your own projects.
```
Josh-Schifter, line 28: consult

src/frontend-samples/snow-decoration-sample/README.md
```diff
@@ -0,0 +1,28 @@
+# Particle Effect - Snow & Rain
```
Josh-Schifter, line 1: This readme should probably describe the difference between rain (static) and rain (dynamic)

Josh-Schifter (commented):
Maybe add a 'pause' button.  I found it difficult to read the code while the snow is moving in my peripheral vision.

Ellord207:
@pmconne, if you have no more outstanding changes, please approve or review.  I would like for this to be merged before release day tomorrow.

src/frontend-samples/fire-decoration-sample/FireDecorator.ts
```diff
@@ -13,7 +13,7 @@ function randomInteger(min: number, max: number): number {
   return Math.floor(Math.random() * (max - min + 1)) + min;
 }
 
-/** Generate random floating-point number in [min, max). */
+/** Generate random floating-point number in [min, max]. */
```
pmconne, line 16: This was expressed as `[min, max)` because the range includes min but exclude max. Did you change that behavior?

pmconne:
The snow falls unrealistically slowly.

pmconne:
The rain particles are very difficult to see against the sky color.

pmconne:
I see small (1 pixel) snow particles that zip down the screen very quickly (maybe existing only 2-4 frames).

src/frontend-samples/fire-decoration-sample/FireDecorator.ts
Ellord207: Was unaware of this notation.  Reverting change.

Ellord207:
> I see small (1 pixel) snow particles that zip down the screen very quickly (maybe existing only 2-4 frames).

Sounds like an issue with lag.  At that size, it should have its velocity reduced by 40% (or "unrealistically slow").  I watched it for a while and didn't see anything like this without causing lag.  I'm not really sure what to do about this.

Assuming this was run on your massive monitor, perhaps the density grew the particle count higher than intended.

pmconne:
> Sounds like an issue with lag.

No lag. I observe it as soon as I open the sample, with no adjustments to the settings.

pmconne (approved):

Josh-Schifter (approved):

PR #183, opened 2021-04-02
Upgrade 2.14

williamkbentley:
- TileRequestChannel changes
- AccessTokenProps property name changes


pmconne:
@williamkbentley please let me know when this gets merged so I can resurrect #161.

StefanRetief (approved):
Works with iTwin Viewer

src/NoSignInIAuthClient.ts
```diff
@@ -47,12 +47,12 @@ export class NoSignInIAuthClient implements FrontendAuthorizationClient {
 
     const response = await fetch(userURL);
     const body = await response.json();
-    const tokenJson = {
+    const tokenJson: AccessTokenProps = {
       ...await body,
       // eslint-disable-next-line @typescript-eslint/naming-convention
-      _userInfo: { id: "MockId" },
```
BaoTon, line 53: Should getDevAccessToken() also have the same change?
In this case we're setting the authClient for iModelApp which seems to be bound to no underscores now. But for developer.bentley.com, this doesn't seem to be the case.

BaoTon: It seems like APIs from developer.bentley.com still assumes the old tokens with underscores, this should be fine for now.

BaoTon: For uniformity, let's change getDevAccessToken() for this change as well.

williamkbentley: Makes sense

BaoTon (approved):
LGTM thanks!

src/NoSignInIAuthClient.ts
williamkbentley: Was also able to remove the lint `eslint-disable-next-line`

PR #185, opened 2021-04-08
Support for iTwin Viewer

StefanRetief:
Support for the iTwin Viewer and flow for sample widgets.


BaoTon:
Occasionally getting this error when switching samples:

image

BaoTon:
Another issue caused by switching between samples:

image

src/frontend-samples/view-attributes-sample/ViewAttributesWidget.tsx
```diff
@@ -0,0 +1,190 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import "common/samples-common.scss";
```
BaoTon, line 5: These common/samples-common.scss no longer needs to be imported right?

BaoTon (commented):
I think the WebGL Context is still not being properly cleaned up. Ocassionally getting memory issues:

image

Solved with close on IModelConnections

williamkbentley:
I got this error when opening GlobalData sample
image

williamkbentley (approved):
Looks good to me.

I got WebGl error only once when I was changing samples rapidly. Like beyond reason rapidly.

BaoTon:
> Looks good to me.
> 
> I got WebGl error only once when I was changing samples rapidly. Like beyond reason rapidly.

Same here, that's the biggest concern I have but else it also looks good on my end.

PR #188, opened 2021-04-14
Resizing fixes

StefanRetief:
Widgets:
- Added listener to adjust floating widgets when the viewport resizes.

Opening/Closing panes
- Removed transitions
- Fixed issue with buttons being in the wrong state.

The issue with the transitions on the pane is that the onChange handler from the split-screen immediately calls the callback for adjusting the size. Since there is a transition, the width of the pane being sent to the handler is still 0, even though it's currently going through a transition to the correct width. I will need to add some additional logic to the panes before transitions would be fully supported.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2190,12 +2190,12 @@
       },
       "dependencies": {
         "@typescript-eslint/eslint-plugin": {
-          "version": "4.21.0",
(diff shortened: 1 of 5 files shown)


Josh-Schifter (approved):
When I collapse or expand the editor pane there is a moment where the widgets hop to the left and then back again.  Same with other types of resize.  It looks unprofessional.  I'm approving because its an improvement but please try to address the hopping.

PR #189, opened 2021-04-16
State reset errors

StefanRetief:
So there's some fun stuff in this PR.

Basically, I removed the abort controller because it wasn't quite getting the job done and I removed the debouncer. The debouncer was only delaying the inevitable, vs fixing it.

Essentially, what I've found is that the errors relating to the state not being set up properly for the viewer lie in the iTwin Viewer Initializer. Essentially, what is happening is that as the viewer gets swapped out every time we rapidly switch the sample, the initializer is updating a property called "_initializing". What happens when we switch the samples is sometimes the Initializer sticks around and we are stuck with the Initializer stuck in the "initializing" state. This causes a big issue because now every viewer rendered after this "failed" viewer will also fail to render because it hasn't been initialized properly. 

I've added a hack in this PR to fix it, but the Viewer team will need to fix it as well. It isn't perfect because we will still have errors in the console, but the sample is essentially self-correcting and will continue to function as it should, regardless of the errors in the console.


Ellord207:
Still able to reproduce issue.

StefanRetief:
> Still able to reproduce issue.

Crashing or just a console message?

Ellord207:
> > Still able to reproduce issue.
> 
> Crashing or just a console message?

"ValueRendererManager.js:91 Uncaught (in promise) Error: PropertyValueRendererManager.registerRenderer error: type 'SelectableInstance' already registered to 'InstanceKeyValueRenderer'"
And crash

BaoTon:
Looks like I managed to get a different error:

image

I tried replicating Nathan's issue but got this instead of swapping the samples at a very precise time. Swapping between 3d and 2d sample.

BaoTon:
Also this issue:

image

I think it's easier to spot these issues if you throttle the Network speed. I have mine set to "Fast 3G"

StefanRetief:
> 
> 
> Also this issue:
> 
> image
> 
> I think it's easier to spot these issues if you throttle the Network speed. I have mine set to "Fast 3G"

So these errors are related to promises in the sample themselves. There's unfortunately no way to cancel these without adding a bunch of extra code to the samples themselves. However, I did notice that when running a production build, these would only sit in the console, there didn't seem to be any side effect, though I haven't looked into what it could mean for the memory usage of the app

BaoTon:
This is a new one that doesn't seem to be recoverable:

image

All samples in Viewer Features broke after this.

StefanRetief:
> 
> 
> This is a new one that doesn't seem to be recoverable:
> 
> All samples in Viewer Features broke after this. I can get this pretty consistently swapping between 3d sample and camera path.
> This seems to happen everytime.

My latest push should address this issue.

BaoTon:
Still getting IModelAppl.startup issues:
image

An error is no longer obvious and just shows the blank grey screen. SelectableInstance error is also occasionally occurring - but again showing in only console.

StefanRetief:
> Looks like the .then issues disappeared. Still getting IModelAppl.startup issues:
> 
> 
> An error is no longer obvious and just shows the blank grey screen. Seems to be random when just swapping samples. SelectableInstance error is also occasionally occurring - but again showing in only console.

Yes, that's expected. Waiting on the viewer PR (https://github.com/imodeljs/itwin-viewer/pull/98) to be completed and it should resolve the gray screen.

BaoTon (approved):
LGTM

PR #195, opened 2021-04-21
Add hilite ruleset. It gets removed on re-init of presentation

williamkbentley:


src/@itwinjs-sandbox/hooks/useSampleWidget.ts
```diff
@@ -36,6 +39,21 @@ export const useSampleWidget = (instructions: string, iModelList?: SampleIModels
...
+    const addHiliteRuleset = async () => {
+      try {
+        if (Presentation.presentation)
+          await Presentation.presentation.rulesets().add((HILITE_RULESET as any).default as Ruleset);
+      } catch {
+        // Presentation not initialized
+      }
+    };
+    addHiliteRuleset();
+  });
```
StefanRetief, line 55: I suggest adding [] as your dependencies on this useEffect, otherwise this will get called on every state change

src/Components/SampleVisualizer/SampleVisualizer.tsx
```diff
@@ -79,6 +81,9 @@ export const SampleVisualizer: FunctionComponent<SampleVisualizerProps> = ({ iTw
           }
 
           setAppReady(true);
+
+          // Fix to add Hilite ruleset until Bug #599922 is addressed
+          await Presentation.presentation.rulesets().add((HILITE_RULESET as any).default as Ruleset);
```
StefanRetief, line 86: It may be a good idea to move this before the setAppReady(true). It may run into a race condition with rendering the sample otherwise

williamkbentley: Ok. I'll also add it in the `!iTwinViewerReady` if block since iTwinViewer samples will get this through the hook

StefanRetief (approved):

BaoTon (approved):

PR #197, opened 2021-04-26
Modified viewSetup to check iModel Connection & modifed floating widgets

StefanRetief:
Added a check to the imodel connections when creating a viewstate. This often resulkted in a crash when switching samples.

Modified how the floating widgets are calculated. Previously the widgets were rendered and unrendered before checking the size, resulting in the widgets throwing errors. This method wraps the react node and measures the widgets as they are rendered for the sample.


Ellord207 (approved):

dustinlebsock (approved):

BaoTon (approved):
LGTM

PR #202, opened 2021-05-07
Multi-viewport Sample iTwinViewer update

dustinlebsock:


Ellord207 (changes requested):
1. I'm not seeing two viewports.
2. The FrontstageProvider file is not included in the editor.  Is there a reason for this?
3. Visible edges are turned on.  Are you getting the default view from the ViewSetup?

dustinlebsock:
> 1. I'm not seeing two viewports.
> 2. The FrontstageProvider file is not included in the editor.  Is there a reason for this?
> 3. Visible edges are turned on.  Are you getting the default view from the ViewSetup?

For points 1/3, I never ran into these issues. Point 2 has been addressed.

There was a bug I noticed when switching iModels from the retail building to the metrostation would cause metrostation to not update the viewport. And the House imodel's viewport would update at all. Both of these issue have been addressed. See if they solve what you are talking about.

Ellord207 (approved):
> For points 1/3, I never ran into these issues.

After pulling and rebuild, the issues went away. 🤷‍♀️  Was unable to reproduce.

I am curious what happened to the original css styles for the selected viewports.  I'm assuming the red line is from the ui-framework and should be prioritized over the old styles from the UX team.

PR #203, opened 2021-05-08
fixed issues with split pane not working on smaller screen

StefanRetief:


StefanRetief:
> The expand button is not working for me with screen sizes less than 1000 px wide. The pane does not display.

Which pane? On screens sized smaller than 768px, the editor pane should be disabled since the editor, preview and gallery cannot exist at the same time.

On screen sizes less that 1024px, only one pane can open at a time.

Ellord207:
For me, I can't open the gallery if the window is less than 994px wide.

I know this would be complicated, but especially in the case of a mobile environment, vertically stacking the editor and code maybe a good option.

StefanRetief:
> 
> 
> For me, I can't open the gallery if the window is less than 994px wide.
> 
> I know this would be complicated, but especially in the case of a mobile environment, vertically stacking the editor and code maybe a good option.

I'm thinking that in screen sizes less than 1024px but greater than 768px, we can keep the same behavior of having different panes on the sides, but limited to one pane at a time. I think decreasing the min-size on the preview would also be beneficial.

For screen sizes less than 768px, it may make sense to allow the editor and sample gallery to expand over the entire screen when opened, as though they were overlays on top the of preview.

Ellord207 (changes requested):
1. The editor and gallery panels are both hidden by default.
2. It would be good to see the gallery and editor open buttons when the other is the only one shown.

PR #204, opened 2021-05-10
Explode itwinviewer

dustinlebsock:
Updates the Explode Sample to use the iTwinViewer component.


src/frontend-samples/explode-sample/ExplodeWidget.tsx
```diff
@@ -0,0 +1,193 @@
+import React, { useEffect } from "react";
```
williamkbentley, line 1: Copyright

williamkbentley (changes requested):
Just copyright missing. I'm not sure what the Zoom To adds to this sample, but that's a different discussion

williamkbentley (approved):

Ellord207 (approved):
While checked if the README was outdated, I found some spelling issues both new and old.  Easier to just fix myself since I was fixing the old anyways.  Looks good.

PR #207, opened 2021-05-13
Add index component

BaoTon:
Added a new "SearchIndex" component that populates from sampleManifest.ts.

- Added backdoor to the SPA if search params match a "?getSearchIndex" that renders "SearchIndex" component
- The index is constructed from "sampleSpec.ts" per sample
- "?getSearchIndex" page will be scraped from a backend indexer/builder.

Note that this is the most unintrusive way I could generate the index without doing a refactor of the existing code. 
If there is a way to generate a "samplesIndex.json" post-build and dump it into the builds directory, this would be preferable.

I tried using ts-node and running "sampleManifest.ts" independently but this requires several configurations to sampleSpec and tsconfig due to the dynamic loader.

Considering how clean this implementation is, I'm not sure if that would be any better depending on how much the repo must change.

### DIFF
--- a/src/Components/App/App.tsx
+++ b/src/Components/App/App.tsx
@@ -5,7 +5,7 @@
 
 import React, { FunctionComponent } from "react";
 import { SampleShowcase } from "Components/SampleShowcase/SampleShowcase";
-import SearchIndex from "SearchIndex";
+import SearchIndex from "Components/Search/SearchIndex";
 
 export const App: FunctionComponent = () => {
 
--- a/src/Components/App/App.tsx
+++ b/src/Components/App/App.tsx
@@ -5,8 +5,16 @@
 
 import React, { FunctionComponent } from "react";
 import { SampleShowcase } from "Components/SampleShowcase/SampleShowcase";
+import SearchIndex from "SearchIndex";
 
 export const App: FunctionComponent = () => {
+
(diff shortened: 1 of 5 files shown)


williamkbentley:
So you're just indexing name and group?

BaoTon:
> So you're just indexing name and group?

Currently just name and group name since these are the only attributes that makes sense in sampleSpec.ts. 
We could add label maybe? README.md is way too much content.

I think adding a "tags" and "short description" might improve it.

Josh-Schifter:
Just like sandboxes, each showcase sample should have a short description that contains tags.  We should index those tags.

BaoTon:
I assume we'll want to wait for all samples to have a new attribute called "shortDesc", which includes a description and optional tag before we can merge this then.

Josh-Schifter:
We should write a few short descriptions but no need to wait for them all, right?  If the description is missing, then there are no tags to add to the index.

BaoTon:
Yeah I can do that. For now I won't index the contents README.md. I'll add a new index for short description and moving forward we can add these in our sampleSpec.ts.

williamkbentley (approved):
Do we need to add some descriptions in separate PRs?

BaoTon:
> Do we need to add some descriptions in separate PRs?

Yep, we'll need to slowly add descriptions as we go if we want them indexed.

PR #216, opened 2021-05-24
v2.16. Removed trivial @internal API usage

williamkbentley:
- Update to v2.16
- Remove some @internal API usage


BaoTon (approved):
Build looks good - all samples affected looks good.

LGTM.

PR #219, opened 2021-06-01
Annotations

StefanRetief:
Added ability to start generating annotations. Annotations rely on two comments:
- `START ANNOTATION [0-9]+`
- `END ANNOTATION [0-9]+`

Using a webpack loader, the annotations will be generated around (not inclusive) the code specified. For example, if the desired code is on line 4-8, the comments should be placed on line 3 and 9. Upon compilation, the comments will not be removed from the code, but will be hidden within the editor on local host.

Once the annotations module of the editor is complete, the markdown should be fed to a react component that will parse out the steps and annotations as needed.


config-overrides.js
```diff
@@ -26,6 +27,18 @@ module.exports = function (config, env) {
...
   }
 
+  config.module.rules.unshift({
+    test: /frontend-samples.*\.(ts|tsx|scss)$/,
+    enforce: "post",
+    loader: path.resolve("./annotation-loader/index.js"),
+    options: {
+      start: /(?:START)\s*ANNOTATION\s*[0-9]+/i,
+      end: /(?:END)\s*ANNOTATION\s*[0-9]+/i,
+      identifier: /[0-9]+/i,
```
williamkbentley, line 37: Maybe instead of `[0-9]+` we could use `\d+\.?\d*`. That way they can be decimals as well. Helps in the case of wanting to add an annotation between 2 and 3 for example. The author would be able to use 2.1 instead of bumping everything after three.

annotation-loader/GetOptions.js
```diff
@@ -0,0 +1,55 @@
+const getLoaderOptions = require('loader-utils').getOptions
```
williamkbentley, line 1: Copyright

annotation-loader/ParseAnnotationLocations.js
```diff
@@ -0,0 +1,87 @@
+const rawLoaderRegex = /export default \"(.*)\"/i;
```
williamkbentley, line 1: Copyright

config-overrides.js
```diff
@@ -3,10 +3,11 @@
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 const MonacoWebpackPlugin = require("monaco-editor-webpack-plugin");
-
+const path = require("path");
 const {
   override,
   addWebpackPlugin,
+  // addWebpackModuleRule,
```
williamkbentley, line 10:
```suggestion

```

config-overrides.js
Josh-Schifter: Might be better to allow arbitrary alphanumeric tags.  Putting the sequence into the annotations themselves will be somewhat limiting.  I'm assuming the walkthrough sequence will be driven directly from the walkthrough rather than the number of the annotation.

StefanRetief: For reference, the reality data sample is converted over to use this method in this PR. Maybe that would give you a better idea of how it would look.

src/annotation-loader.d.ts
```diff
@@ -0,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+export interface AnnotationStep {
```
StefanRetief, line 5: This is temporary until the editor module is completed

src/frontend-samples/reality-data-sample/sampleSpec.ts
```diff
@@ -12,12 +12,13 @@ export function getRealityDataSpec(): SampleSpec {
     label: "Reality Data",
     image: "reality-data-thumbnail.png",
     iTwinViewerReady: true,
-    readme: async () => import("!!raw-loader!./readme.md"),
+    readme: async () => import("!raw-loader!./readme.md"),
+    walkthrough: async () => import("!annotation-loader!./annotations.md"),
```
StefanRetief, line 16: This will create a json object array from this file that can be consumed by the annotations module

src/frontend-samples/reality-data-sample/sampleSpec.ts
```diff
@@ -12,12 +12,13 @@ export function getRealityDataSpec(): SampleSpec {
...
     files: () => [
-      { name: "RealityDataApi.ts", import: import("!!raw-loader!./RealityDataApi") },
-      { name: "RealityDataApp.tsx", import: import("!!raw-loader!./RealityDataApp"), entry: true },
-      { name: "RealityDataWidget.tsx", import: import("!!raw-loader!./RealityDataWidget") },
-      { name: "RealityData.scss", import: import("!!raw-loader!./RealityData.scss") },
+      { name: "RealityDataApi.ts", import: import("!annotation-raw-loader!./RealityDataApi") },
```
StefanRetief, line 18: The "annotation-raw-loader" will strip the annotation comments from the file and load the file the raw-loader does.

williamkbentley:
We should probably add some documentation in CONTRIBUTING or README

StefanRetief:
> 
> 
> We should probably add some documentation in CONTRIBUTING or README

Added to CONTRIBUTING

williamkbentley (approved):

PR #228, opened 2021-06-11
Walkthrough components

StefanRetief:
Added Annotation components.

To Test: By default, the walkthrough will be disabled until the key in launch darkly is enabled. The override for this is to add a url parameter like so: ```http://localhost:3000/?sandbox-walkthrough=true```

Only a handful of samples have walkthroughs, but the most complete ones are the markerpin sample and the 3D viewer sample.


StefanRetief:
@linasburneika I had to include some type imports for the ```AnnotationStep``` interface. I just wanted to make sure it doesn't bloat the app again.

BaoTon (approved):
LGTM

PR #229, opened 2021-06-16
Refactor featuretoggle and include url overrides

williamkbentley:


src/@itwinjs-sandbox/hooks/useSampleIModelParameter.ts
```diff
@@ -33,18 +33,6 @@ const updateiModelParam = (imodel?: string, context?: string) => {
...
-
-  // Detect if editor was enabled in URL params as a semi-backdoor, this
-  // bypasses the ld feature flag
-  const editorEnabled = new URLSearchParams(window.location.search).get("editor");
-  if (editorEnabled) params.append("editor", editorEnabled);
-
-  window.history.pushState(null, "", `?${params.toString()}`);
-
-  // Send to parent if within an iframe.
-  if (window.self !== window.top) {
```
StefanRetief, line 45: I think this is probably needed for the current iframe view we have set up? I didn't see it moved anywhere else

williamkbentley: I dont see where the editor flag is used elsewhere in the code. Can you point me to where the editor is actually hidden by this flag?

BaoTon: This isn't related to the editor flag. This pushes search parameters up to the parent iframe. You only need to remove lines 39/40.

StefanRetief: Sorry, should have been more specific. I meant that lines 45-46 are probably necessary else where. Without `postMessage`, we never get up updated address bar when the showcase is in an iframe.

williamkbentley: I think this also happens in ActiveSample. But doesnt hurt to keep

williamkbentley:
@BaoTon @StefanRetief Mind re-reviewing?

StefanRetief (approved):
LGTM

BaoTon (approved):
LGTM

PR #234, opened 2021-06-28
Editor file loader

StefanRetief:
Modified the editor file loader to parse out the file name. Previously, developers needed to specify file names in the sample spec as well as import. This resulted in cases where the file names in the sample spec didn't match that in the file structure, leading to issues when referencing in either the readme or walkthroughs.


williamkbentley (approved):
I can't find any issues with this one. Feels snappier, if anything. Nice

** Pending CI build pass

PR #245, opened 2021-07-13
Version compare Update

Ellord207:
Updating after feedback while on QA.


dustinlebsock:
You can remove the 'import "common/samples-common.scss";' line in changeelementswidget.tsx. But it looks good!

dustinlebsock (approved):

williamkbentley (approved):
It doesn't really make sense to compare one version to itself (i.e "Final Version" to "Final Version"). But that is probably an implementation detail.

PR #246, opened 2021-07-19
Element Change: Feedback from ecosystem team

Ellord207:
Added toast message for the number of elements returned by API.
Emphasis no longer removed on no changes.
Added more explanation in sample information.


Josh-Schifter (approved):

BaoTon:
Probably not an issue for this PR but does APIM requests not cache? 

The difference between the final version changeset and the first spatial changeset is over 10x larger.
Request for final version is 196 Bytes.
Request for first spatial version is 602 KB.

Maybe demonstrating a caching mechanism might be a good idea - swapping between them should be instantaneous after the first request.

PR #248, opened 2021-07-22
Add filtering to issues

dustinlebsock:
Two drop down menus created to allow filtering of issues.


src/frontend-samples/issues-sample/IssuesWidget.tsx
```diff
@@ -454,19 +484,39 @@ const IssuesWidget: React.FunctionComponent = () => {
     ));
   };
 
+  const filterTypes = ["All", "Civil Design", "Clash", "Closeout", "Communication", "Deficiency", "Design", "Field Data", "Issue", "Observation", "Other", "Punchlist", "Risk", "RFI", "Submittal Issue", "Task", "Transmittal Issue"];
```
Josh-Schifter, line 487: For several of these filter types, there are no matching issues in the sample iModel.  Why do we need so many in the picker?  It might make sense if they list was being retrieved from some API, but if we are hardcoding the list, I suggest just leaving out most or all of the ones with no issues.

src/frontend-samples/issues-sample/IssuesWidget.tsx
```diff
@@ -454,19 +484,39 @@ const IssuesWidget: React.FunctionComponent = () => {
...
     <>
       <div style={{ height: "300px" }}>
+        {/** Only display header when issue isn't selected */}
+        {!currentIssue &&
+          <div className="issue-list-header">
+            <Subheading style={{ margin: "0", padding: "8px 5px", color: "#fff" }}>{`Issues (${currentIssues.length})`}</Subheading>
+            <div className="issue-list-filters">
+              <div className="filter">
+                <span>State</span>
+                <Select options={["All", "Closed", "Open", "Draft"]} value={stateFilter} onChange={(option) => setStateFilter(option.target.selectedOptions[0].value)} />
```
Josh-Schifter, line 499: I suggest we create at least one closed issue in the sample iModel to make this picker do something in the showcase.

Josh-Schifter:
When I click on the first issue (FID-00003) the view changes and I can see several markers, but none of them are FID-00003.

image

dustinlebsock:
> When I click on the first issue (FID-00003) the view changes and I can see several markers, but none of them are FID-00003.
> 
> image

The issue was attached to the steal beam and what ended up happening is the marker was displayed way behind the camera. A new issue has been attached where this doesn't happen.

src/frontend-samples/issues-sample/IssuesWidget.tsx
dustinlebsock: Only the filters that return a result are displayed now.

src/frontend-samples/issues-sample/IssuesWidget.tsx
dustinlebsock: A closed issue has been created! Thanks @Josh-Schifter

BaoTon:
Can we call applyView() when selecting the issue in the widget too?

BaoTon:
I did not know how to get back to the list of issues for the longest time.

image

I think a simple solution is just enable the hover effect by default.

image

dustinlebsock:
> Can we call applyView() when selecting the issue in the widget too? I noticed the camera only changes when we click on the decorator - would be nice if it changed when focusing on an issue.

I don't like the idea of zooming to on issue select. To me it seems a user just wants to view the content/response. If they wanted to change the view, they would click the zoomTo magnifying glass on the issue. For the marker it makes sense though because it matches all the other samples, and the user isn't sure what issue its actually pointing to.

src/frontend-samples/issues-sample/Issues.scss
```diff
@@ -23,6 +23,20 @@
 .issue-list-header {
     padding: 6px;
     border-bottom: 1px solid $buic-background-divider;
+    display: flex;
+    flex-direction: column;
```
BaoTon, line 27: Could we change flex-direction here to "row" and maybe add some margin? 
I think there is a lot of wasted space in the widget and would benefit if they were in the same line.

image

BaoTon: On second thought, default size is pretty tiny so might just be okay if they're on separate lines.

BaoTon (approved):
Code LGTM.

Minor stylistic preferences but I don't think any are showstoppers. I'll leave it up to you if you want to address them or not.

dustinlebsock:
> I did not know how to get back to the list of issues for the longest time.
> 
> image
> 
> I think a simple solution is just enable the hover effect by default.
> 
> image

Made it more noticeable with a border.
image

PR #249, opened 2021-07-23
Tree sample support

StefanRetief:
Added support for a viewer without an iModel viewport. Instead, given a component that accepts an imodel connection, you can now make samples that can rely on the imodel connection provided by the viewer and use it for the main content in the display. 

Currently, all UI Tree samples use this method.

Also included in this PR are the @itwinjs-sandbox files being exposed in the editor. Right clicking "Go To Definition" will reveal the contents of the file in @itwinjs-sandbox instead of just the minified typedef.


roluk:
When attempting to edit tree samples, I'm getting error `IModelApp.startup must be called before calling Presentation.initialize`.

StefanRetief:
> When attempting to edit tree samples, I'm getting error `IModelApp.startup must be called before calling Presentation.initialize`.

I added some retry logic to make sure the sample will try again when it fails to load correctly. 

An unfortunate part of how the samples work is that since the initialization relies on async calls and there isn't a good way to check he status of the call, or where the sample is in the rendering process, these errors are bound to happen. More or less, we want to minimize the impact on users, so if an error pops up in the console but the viewport is still in a useable state (barring having to close the react error handling page), that is all we can do.

BaoTon (approved):
LGTM, address lint issues and undo Custom Table Node Tree and you should be good to go

PR #251, opened 2021-08-05
Necessary adjustments for the new Metrostation model

williamkbentley:
- Serialized view - Updated with new ids
- Serialized view - Added new relevant views
- Camera path update with new path
- Some view setup changes for new model (It lacks a decent default view)
- Minor Camera path enhancements


williamkbentley:
TODO: Elevation of Issue placement and views needs to be adjusted. The new Metrostation model is 3 meters higher than the old one.

Ellord207 (commented):
Is there a reason that you only updated the ViewSetup file in the @itwinjs-sandbox package, and not also the one under api?

williamkbentley:
> Is there a reason that you only updated the ViewSetup file in the @itwinjs-sandbox package, and not also the one under api?

I thought the one under api was for legacy (non iTwin-viewer) based samples, of which we have none.

Ellord207:
> I thought the one under api was for legacy (non iTwin-viewer) based samples, of which we have none.

Okay, I assumed the file would have been deleted with the last of the legacy samples.

williamkbentley:
> > I thought the one under api was for legacy (non iTwin-viewer) based samples, of which we have none.
> 
> Okay, I assumed the file would have been deleted with the last of the legacy samples.

Yeah, possibly it was overlooked. We can address that in a separate PR

BaoTon:
The widget appears to always be disabled for serialize view sample:

image

williamkbentley:
> The widget appears to always be disabled for serialize view sample:
> 
> image

You need to edit SampleViewStates.ts. Change iModelId to `b55dec38-d9b7-4029-9b9c-6b899151328f` for Metrostation Sample

BaoTon (approved):
LGTM

PR #272, opened 2021-08-31
Dl/camera path stop motion fix

dustinlebsock:


src/frontend-samples/camera-path-sample/CameraPathWidget.tsx
```diff
@@ -170,8 +170,8 @@ const CameraPathWidget: React.FunctionComponent = () => {
   const _onChangeRenderSpeed = (currSpeed: string) => {
     let speedOfMotion: number = 0;
     switch (currSpeed) {
-      case "5 Mph: Walk":
-        speedOfMotion = 2.2352; // 4mph = 1.788 meters/second
+      case "5 Mph: Walking":
```
williamkbentley, line 173: The initial `setSpeedLevel` sets it to "3 Mph: Walking". Maybe these options should be an enum or something?

dustinlebsock: refactored. Made a list that tightly couples the text and m/s together. See if that looks better

williamkbentley (approved):

PR #273, opened 2021-09-03
Trim down to one client

williamkbentley:


williamkbentley:
> I'm seeing an error like this for every sample that tries to open an iModel. Not only MetroStation, but other iModels as well.
> 
> Error: Project with name "Metrostation Sample" does not exist
> at getIModelInfo (useSampleIModelConnection.ts:34)
> at async useSampleIModelConnection.ts:96

I had to complete a corresponding backend deployment. Should be working now. Can you try again?

Josh-Schifter:
Switching iModels is acting a little flaky but I can reproduce on master.  So this PR seems good.

Josh-Schifter (approved):

PR #276, opened 2021-09-09
Issue Markers

tomdicarlo:
Resolved a bug caused by a race condition between the initial view setup and the creation of the marker pin decorator that caused markers to not render correctly. MarkerSets would be created before the initial view was setup, which required them to be updated manually.

Resolved another bug, also caused by a race condition, that caused the markers in the issue sample to not update on the very first decoration frame after having either of the filters change. The fix is hacky, it uses a timer to make a call to invalidate decorations, but this bug seems to fall outside of the sample, since the correct markers are being supplied to the DecorateContext on the first call to decorate.

### DIFF
--- a/src/frontend-samples/issues-sample/IssuesApi.ts
+++ b/src/frontend-samples/issues-sample/IssuesApi.ts
@@ -55,6 +55,5 @@ export default class IssuesApi {
 
   public static clearDecoratorPoints(decorator: MarkerPinDecorator) {
     decorator.clearMarkers();
-
   }
 }
--- a/src/frontend-samples/issues-sample/IssuesApi.ts
+++ b/src/frontend-samples/issues-sample/IssuesApi.ts
@@ -15,7 +15,6 @@ export interface LabelWithId extends LabelDefinition {
 }
 
 export default class IssuesApi {
-  public static _issuesPinDecorator?: MarkerPinDecorator;
 
   public static async getElementKeySet(elementsId: string) {
     if (!elementsId || elementsId.trim().length === 0)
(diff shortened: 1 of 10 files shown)


src/frontend-samples/issues-sample/IssuesWidget.tsx
```diff
@@ -189,7 +193,12 @@ const IssuesWidget: React.FunctionComponent = () => {
         setCurrentIssue(iss);
       });
     }
-  }, [applyView, currentIssues]);
+    /** This timer is being used to resolve an async issue that prevents marker pin decorations from being drawn immeditately after changing the current issues */
+    const timer = new Timer(250);
+    timer.setOnExecute(() => { if (IModelApp.viewManager) IModelApp.viewManager.invalidateDecorationsAllViews(); });
+    timer.start();
```
dustinlebsock, line 199: Really not a fan of this and there has to be a better way. We just aren't doing something right. I've played around with it for a few minutes but i couldn't solve it trivially. It's like its just not rendering the next frame after changing the decoration. Maybe we should reach out to someone for help.

src/frontend-samples/issues-sample/IssuesApi.ts
```diff
@@ -36,37 +35,25 @@ export default class IssuesApi {
...
   }
 
-  public static decoratorIsSetup() {
-    return (null != this._issuesPinDecorator);
-  }
-
   public static setupDecorator() {
-    if (undefined === this._issuesPinDecorator)
-      this._issuesPinDecorator = new MarkerPinDecorator();
+    return new MarkerPinDecorator();
```
dustinlebsock, line 39: You can just make this a static variable that's instantly initialized. Its never set or deleted so a lot of code can be removed then.

src/frontend-samples/issues-sample/IssuesApi.ts
```diff
@@ -36,37 +35,25 @@ export default class IssuesApi {
...
   }
 
-  public static addDecoratorPoint(issue: IssueGet, pinImage: HTMLImageElement, title?: string, description?: string, onMouseButtonCallback?: any) {
+  public static addDecoratorPoint(decorator: MarkerPinDecorator, issue: IssueGet, pinImage: HTMLImageElement, title?: string, description?: string, onMouseButtonCallback?: any) {
```
dustinlebsock, line 42: the decorator param can be removed because it can just rely on the static decorator. The same idea will be propogated through the rest of the methods.

src/frontend-samples/issues-sample/IssuesWidget.tsx
```diff
@@ -41,17 +43,19 @@ const IssuesWidget: React.FunctionComponent = () => {
   const [stateFilter, setStateFilter] = useState<string>("All");
   /** The type filter. */
   const [typeFilter, setTypeFilter] = useState<string>("All");
+  /** The decorator used for displaying issue markers */
+  const [issueDecorator] = React.useState<MarkerPinDecorator>(() => {
+    const decorator = IssuesApi.setupDecorator();
+    IssuesApi.enableDecorations(decorator);
+    return decorator;
+  });
```
dustinlebsock, line 51: This can just be removed because the decorator will already be initialized.

src/frontend-samples/issues-sample/IssuesWidget.tsx
```diff
@@ -41,17 +43,19 @@ const IssuesWidget: React.FunctionComponent = () => {
...
 
   /** Initialize Decorator */
   useEffect(() => {
-    IssuesApi.setupDecorator();
-    IssuesApi.enableDecorations();
```
dustinlebsock, line 48: this will need to be readded though.

src/frontend-samples/issues-sample/IssuesApi.ts
tomdicarlo: I think that could be a potential race condition issue for the teardown. Right now, two instances of the widget are being created for each of the samples, one of which is torn down right away. That could potentially tear down the static decorator after the "real" widget sets it up. I could add some sort of check to figure out which widget is actually being used, but I feel like that is adding more complication than just having the widget maintain the decorator.

dustinlebsock: I'm not following, the decorator would never be deallocated or torn down if its a static variable. If `disabledecorations()` is the problem, then it would still be a problem with the current code as it would just leave the 'old' decorator in viewmanager before rendering the second widget.

tomdicarlo: Sorry I could see how my wording was confusing above. But yes, disableDecorations() is the problem if we went with a static approach, since the widget that is torn down in the beginning will remove the static decorator from the viewManager, potentially after the other widget has attempted to add it to the viewManager. But with the current approach, both widgets are creating their own decorator, so the widget that is torn down will only remove it's own decorator from the viewManager. The widget that remains for the duration of the sample wouldn't be impacted at all. The only downside of this I could see would be that for a brief period, there could be two versions of the marker pin decorator registered with the viewManager.

Josh-Schifter (approved):

PR #277, opened 2021-09-20
Wkb/point cloud

williamkbentley:
Reality Data Sample
- Added support for multiple reality models
- Point cloud added to Exton campus
- Simplified Reality Data Widget

Other
- Upgraded iModel.js to latest patch version - Necessary for point cloud
- Upgraded to v4 @bentley/react-scripts - Necessary for iModel.js patch version v2.19.12
- react-scripts variables
- Lint


BaoTon (approved):
LGTM

PR #278, opened 2021-09-20
Scientific viz improvements

Josh-Schifter:
Hide the default timeline which a) interferes with the showcase floating widgets and b) confusingly animates even for the static analysis styles.  Now the samples shows how to control the analysis fraction yourself which I think is an improvement anyway.

Also, I added a bunch of code comments.


src/frontend-samples/scientific-visualization-sample/ScientificVizApi.ts
```diff
@@ -160,4 +197,67 @@ export default class ScientificVizApi {
     ScientificVizApi.populateAnalysisStyles(mesh, displacementScale);
     return mesh;
   }
+
+  /** This method shows how to determine if the current analysis style can be animated.  This information is
+   * known when the style is created so it would be reasonable to determine it then.  Here we check after
+   * the fact just to show that it can be done.
+   */
+  public static currentStyleSupportsAnimation(viewport: Viewport) {
```
pmconne, line 205: An analysis style can be animated without having a displacement channel. The scalar channel can be used to change the UV coords used to apply the gradient to the mesh.
Isn't the answer to "can style be animated" just "does it have more than one input"?

Josh-Schifter: >>Isn't the answer to "can style be animated" just "does it have more than one input"?

I don't understand the comment.  Look at the implementation of the method below.  It does NOT look for a displacement channel.  It tests if the style refers to a channel with more than one input.  "1 < channel.data.length".

pmconne: You absolutely look for a displacement channel.
`const channelName = style.displacement?.channelName;`
You only return true if a displacement channel exists.
```
    const channel = auxdata?.channels.find((c) => c.name === channelName);
    let isAnimated = false;
    if (channel)
      isAnimated = 1 < channel.data.length;
```

Josh-Schifter: >You absolutely do look for a displacement channel.

Indeed.  My mistake.

In my limited experience, a channel can have either one or two channels, and both those channels always have the same number of inputs.  Is that true?  If so, I propose this:

```
  public static currentStyleSupportsAnimation(viewport: Viewport) {
    const style = viewport.view.analysisStyle;
    if (!style)
      return false;

    // An analysis style specifies either one or two channelName(s)
    const styleChannels = [style.displacement?.channelName, style.scalar?.channelName];

    // Find either one of those channels in the polyface's auxdata
    const auxdata = ScientificVizDecorator.decorator.mesh.polyface.data.auxData;
    const channel = auxdata?.channels.find((c) => styleChannels.includes(c.name));

    // If the channel has more than one set of data, then the style can be animated by
    // interpolating between each member of the data array.
    let supportsAnimation = false;
    if (channel)
      supportsAnimation = 1 < channel.data.length;

    return supportsAnimation;
  }
```

pmconne: It's not necessarily true that all the channels have the same number of inputs; and a style can specify up to three channels (you left out the normal channel).
Something like:
```
  public static currentStyleSupportsAnimation(viewport: Viewport) {
    const style = viewport.view.analysisStyle;
    if (!style)
      return false;

    const channels = ScientificVizDecorator.decorator.mesh.polyface.data.auxData?.channels;
    const channelNames = [style.displacement?.channelName, style.scalar?.channelName, style.normalChannelName];
    return channels !== undefined && channels.some((c) => c.data.length > 1 && channelNames.includes(c.name));
  }
```

pmconne (approved):

src/frontend-samples/scientific-visualization-sample/ScientificVizApi.ts
pmconne: I thought the same. It wouldn't be very difficult to copy-paste a couple of existing flat mesh styles and omit the displacement in one and the scalar in the other, to demonstrate that they can be used independently. Demonstrating normals would be trickier given your flat mesh will have uniform normals - and you're using an unlit render mode, so normals would have no effect anyway - I wouldn't bother with that.

Josh-Schifter: I think the way to do it would be to rewrite populateAnalysisStyles.  The current pseudo-generalized logic is contrived anyway.  I could replace it with a set of hardcoded styles each which illustrates a different case, ex. displacement only, scalar only, displacement + scalar.  If I did that, I could also give the styles hardcoded names which would be easier to understand and to trace back into the code.  Do you like that?

I happy to  stay away from the normals channel, if only just because of the amount of effort.

pmconne: > Do you like that?

Yes, clearer example = better example.

Josh-Schifter: I took a shot at allowing the user to choose channels so he can mix and match.  I think this makes it easier to understand the relationship between channels and analysisStyle.  Any feedback?

I'm planning to allow the user to choose both a) the color gradient and b) the displacement scale.  It should be pretty easy.

pmconne: Code changes look ok. I can't run it of course.

pmconne: Tried to build + run on my non-dev laptop, since I've been able to do so in the past.
Just in case this is not another case of "webpack hates Paul", are you surprised that I get the following run-time error on startup?
`TypeError: ITwinErrorI18n.use(...).init(...).catch is not a function`

Josh-Schifter: I haven't seen that error myself.

Ellord207 (changes requested):
When changing from an animated displacement to a static one, the slider will move when interacting with drop down controls.

Ellord207:
> > After changing from an actively animated displacement to a static one, the slider will move when interacting with drop down controls.
> 
> Now, I reset the fraction to zero whenever a new channel is selected. Please check if this addresses the issue.

This did not fix the issue.  I'm guessing the issue is that the animator is not getting cancelled when that the analysis style is applied.  You can clear the current animator with ```veiwport.setAnimator(undefined)```.

I would suggest reverting the previous change.  It could create a nice effect changing animation channels to see a smooth transition.

PR #290, opened 2022-01-21
Swiping upgrade

Ellord207:


src/frontend-samples/swiping-viewport/SwipingComparisonApi.ts
```diff
@@ -177,7 +163,7 @@ abstract class SampleTiledGraphicsProvider implements TiledGraphicsProvider {
     viewport.view.forEachTileTreeRef(func);
 
     // Do not apply the view's clip to this provider's graphics - it applies its own (opposite) clip to its graphics.
-    this.viewFlagOverrides.showClipVolume = false;
+    this.viewFlagOverrides.showClipVolume = true;
```
pmconne, line 166: Why did this change from `false` to `true`?

Ellord207: That was an accident.  I do remember it didn't have an effect, as it's redundancy from when the showcase shared a single viewport.

pmconne: I don't follow. From your email description, the swiping sample still divides a single viewport using a clip plane and its inverse.
If you revert this to `false`, rebuild, and run, the same behavior is observed?

> 7.	The TiledGraphicsProvider.addToScene method will
> a.	***Replace the viewports clip with one created using the negated normal (flipping it).***
> b.	Make the reality data transparent
> c.	Create a new Scene context based on the viewport
> d.	Create a new graphic branch and add the graphics created by the scene to it.
> e.	Create graphics from the branch by calling the render system
> f.	Add the graphics to the viewport’s scene.
> g.	***Reset the old clipping plane to the viewport.***
> h.	Make the reality data opaque.

Ellord207: I revented it back to false, and still obvser the same bahavor.

I also pushed the revert.

